PR #25, opened 2022-03-23
feat(server): set up MongoDB connection 

mimoham24:


server/internal/app/repo.go
```diff
@@ -0,0 +1,36 @@
...
+		ctx,
+		options.Client().
+			ApplyURI(conf.DB).
+			SetConnectTimeout(time.Second*10).
+			SetMonitor(otelmongo.NewMonitor()),
+	)
+	if err != nil {
+		log.Fatalf("repo initialization error: %+v\n", err)
+	}
+	if err := mongorepo.InitRepos(ctx, repos, client, "reearth"); err != nil {
```
rot1024, line 31:
```suggestion
	if err := mongorepo.InitRepos(ctx, repos, client, ""); err != nil {
```

server/internal/infrastructure/mongo/mongodoc/util.go
```diff
@@ -0,0 +1,19 @@
...
+		}
+		return f2
+	case bson.M:
+		for _, e := range elements {
+			f2[e.Key] = e.Value
+		}
+		return f2
+	}
+	return f
+}
```
rot1024, line 19: There must be other functions. These will also be used later, so please copy them and comment them out. e.g. `And`

server/internal/usecase/pageinfo.go
```diff
@@ -0,0 +1,63 @@
+package usecase
+
+type PageInfo struct {
+	totalCount      int
+	startCursor     *Cursor
+	endCursor       *Cursor
+	hasNextPage     bool
+	hasPreviousPage bool
+}
```
rot1024, line 9:
```suggestion
type PageInfo struct {
	TotalCount      int
	StartCursor     *Cursor
	EndCursor       *Cursor
	HasNextPage     bool
	HasPreviousPage bool
}
```

This struct contains very little domain logic, so please make all fields public, remove all NewPageInfo functions and methods, and move it to pagination.go.

server/internal/usecase/cursor.go
```diff
@@ -0,0 +1,3 @@
+package usecase
+
+type Cursor string
```
rot1024, line 3: move it to `pagination.go` and delete `cursor.go`

rot1024 (approved):

rot1024 (changes requested):
sorry I approve by mistake!

server/internal/usecase/pagination.go
```diff
@@ -0,0 +1,76 @@
...
+//func NewPagination(first *int, last *int, before *Cursor, after *Cursor) *Pagination {
+//	// Relay-Style Cursor Pagination
+//	// ref: https://www.apollographql.com/docs/react/features/pagination/#relay-style-cursor-pagination
+//	return &Pagination{
+//		Before: before,
+//		After:  after,
+//		First:  first,
+//		Last:   last,
+//	}
+//}
```
rot1024, line 21:
```suggestion
// Pagination is a struct for Relay-Style Cursor Pagination
// ref: https://www.apollographql.com/docs/react/features/pagination/#relay-style-cursor-pagination
type Pagination struct {
	Before *Cursor
	After  *Cursor
	First  *int
	Last   *int
}
```

Remember that public functions in Go do not need to be commented out because it is not an error if they are not used. Nevertheless, this function does not do anything meaningful, so you can delete it.

rot1024 (approved):

PR #30, opened 2022-03-24
feat(server): init echo, config, and gql playground

mimoham24:


server/internal/app/app.go
```diff
@@ -0,0 +1,107 @@
...
+	} else {
+		if ierr := rerror.UnwrapErrInternal(err); ierr != nil {
+			code = http.StatusInternalServerError
+			msg = "internal server error"
+		}
+	}
+
+	return code, msg
+}
+func errorHandler(next func(error, echo.Context)) func(error, echo.Context) {
```
rot1024, line 92:
```suggestion

func errorHandler(next func(error, echo.Context)) func(error, echo.Context) {
```

gofmt

server/internal/app/app.go
```diff
@@ -0,0 +1,107 @@
...
+		))
+	}
+
+	// apis
+	api := e.Group("/api")
+	api.GET("/ping", Ping())
+
+	return e
+}
+func allowedOrigins(cfg *ServerConfig) []string {
```
rot1024, line 53:
```suggestion

func allowedOrigins(cfg *ServerConfig) []string {
```

gofmt

rot1024 (approved):

PR #41, opened 2022-04-08
test(web): introduce vitest

rot1024:
- Introduce vitest instead of jest
- Fix an issue that coverage is not correctly sent to codecov
- Remove slack notification from ci workflow (too noisy)

Ref: https://github.com/vitest-dev/vitest/tree/main/examples/react-testing-lib


KaWaite (approved):
Time for me to study up on testing finally... :P

PR #43, opened 2022-04-12
feat(web): initialize web

issmail-basel:


web/codegen.yml
```diff
@@ -0,0 +1,36 @@
...
+        Cursor: string
+        URL: string
+        Lang: string
+        TranslatedString: '{ [lang in string]?: string } | null'
+        PluginID: string
+        PluginExtensionID: string
+        PropertySchemaID: string
+        PropertySchemaGroupID: string
+        PropertySchemaFieldID: string
+        DatasetSchemaFieldID: string
```
rot1024, line 28:
```suggestion
        DateTime: Date
        FileSize: number
        ID: string
        Cursor: string
        URL: string
        Lang: string
```

web/codegen.yml
```diff
@@ -0,0 +1,36 @@
+overwrite: true
+schema: http://localhost:8080/api/graphql
```
rot1024, line 2:
```suggestion
schema: ../server/schema.graphql
```

The backend is in the same repository, so you can retrieve the schema locally!

web/src/components/atoms/NotFound/index.tsx
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+import { useIntl } from "react-intl";
```
rot1024, line 2: Can you introduce react-i18nnext instead of react-intl?

web/src/main.tsx
```diff
@@ -3,10 +3,13 @@ import ReactDOM from "react-dom";
...
+import loadConfig from "./config";
 
-ReactDOM.render(
-  <React.StrictMode>
-    <App />
-  </React.StrictMode>,
-  document.getElementById("root")
-);
+loadConfig().finally(() => {
+  ReactDOM.render(
```
rot1024, line 9: ReactDOM.render is deprecated in React v18. Use `ReactDOM.createRoot`

rot1024 (changes requested):
Generally OK, but

- Use react-i18nnext instead of react-intl
- Use `ReactDOM.createRoot`

web/src/components/atoms/NotFound/index.tsx
issmail-basel: I removed `react-intl` for now.
Will add `react-i18nnext` it in another issue

rot1024 (approved):

PR #48, opened 2022-04-14
chore(web): introduce i18nnext

issmail-basel:


web/src/i18n.ts
```diff
@@ -0,0 +1,17 @@
+import * as i18n from "i18next";
+import { initReactI18next } from "react-i18next";
+
+import en from "../translations/en.json";
+import de from "../translations/ja.json";
```
rot1024, line 5:
```suggestion
import ja from "../translations/ja.json";
```

rot1024 (approved):

rot1024:
- Add a npm script yarn run i18n to extracts translations and adds  them to the json files
- CI checks all texts are translated (if no, CI fails)
- Change yarn gen:gql to yarn gql because I'm tired of typing every single gen:
- Use yaml instead of json
- Add i18next-browser-languagedetector

web/src/App.tsx
```diff
@@ -3,6 +3,7 @@ import RootPage from "@reearth-cms/components/pages/Authentication/RootPage";
 import { BrowserRouter as Router, useRoutes } from "react-router-dom";
 
 import { Provider as GqlProvider } from "./gql";
+import { Provider as I18nProvider } from "./i18n";
```
KaWaite, line 6: Please remember to use the @reearth-cms alias

KaWaite (approved):
Nice, thanks for keeping it in-line with Inoue-san's PR for reearth-web as well!

PR #53, opened 2022-04-20
feat(web): add authentication 

issmail-basel:


web/src/global.d.ts
```diff
@@ -0,0 +1,10 @@
+import "./config";
+
+declare global {
+  interface Window {
+    REEARTH_E2E_ACCESS_TOKEN?: string;
+    REEARTH_E2E_CESIUM_VIEWER?: Viewer;
+    React?: any;
+    ReactDOM?: any;
```
KaWaite, line 8: Do we need to expose these for CMS? They are exposed in reearth-web now for the extension system, but unless there is a reason for them in CMS, they don't need to be exposed here.

web/src/components/pages/Dashboard/index.tsx
```diff
@@ -0,0 +1,15 @@
+import Typography from "@reearth-cms/components/atoms/Typography";
+import React from "react";
+
+import { AuthenticationRequiredPage } from "../../../auth";
```
KaWaite, line 4: Please use the @reearth-cms alias to avoid import paths like this!

web/src/components/pages/Dashboard/index.tsx
```diff
@@ -0,0 +1,15 @@
+import Typography from "@reearth-cms/components/atoms/Typography";
+import React from "react";
```
KaWaite, line 2: Since we are using Vite and its react plugin, you can stop importing React!

web/src/components/pages/RootPage/index.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { useEffect } from "react";
+import { useNavigate } from "react-router-dom";
+
+import { useAuth } from "../../../auth";
```
KaWaite, line 4: Again, use @reearth-cms alias

web/src/components/pages/RootPage/index.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { useEffect } from "react";
```
KaWaite, line 1:
```suggestion
import { useEffect } from "react";
```

web/src/components/pages/RootPage/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { useNavigate } from "react-router-dom";
+
+import { useAuth } from "../../../auth";
+
+const RootPage: React.FC<{ default?: boolean }> = () => {
```
KaWaite, line 6: Again, phantom "default" prop.

web/src/components/pages/RootPage/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  useEffect(() => {
+    if (isAuthenticated) {
+      navigate("/dashboard");
+    }
+  }, [isAuthenticated, navigate]);
+
+  return isLoading ? (
+    <h1>Loading</h1>
+  ) : !isAuthenticated ? (
+    <>
```
KaWaite, line 19: I know this is boiler code we will replace later, but you don't need <></> here. Change it or don't change it :P

web/src/gql/provider.tsx
```diff
@@ -4,8 +4,10 @@ import {
   ApolloLink,
   InMemoryCache,
 } from "@apollo/client";
+import { setContext } from "@apollo/client/link/context";
 import { onError } from "@apollo/client/link/error";
 
+import { useAuth } from "../auth";
 import { useError } from "../state";
```
KaWaite, line 11: Again, alias.

KaWaite (approved):
Please make the changes I asked for. After that the code looks fine.

KaWaite (approved):
LGTM

Issue #61, opened 2022-04-22
[FE]: Add workspace


issmail-basel:
# What
Related to https://github.com/reearth/reearth-cms/issues/2

# Implementation policy/Detail
- [ ] Fetch workspace data from the backend
- [ ] Add ability to create/read/update/delete workspace
- [ ] Prevent users from deleting or adding members to a personal workspace
- [ ] Add the ability to add users to a workspace (if not personal)
- [ ] User can leave workspace

Issue #35, opened 2022-04-04
[FE]User


mimoham24:
# Description
Front-end development task of #31 

## Todos

- [ ] Fetch user data me query
- [ ] Mutation to change the username

PR #64, opened 2022-05-11
feat(web): add user data and workspace

issmail-basel:
# Related tasks
#35 #61 

## What's done
-  Fetch workspace data from the backend
-  Add ability to create/read/update/delete workspace
-  Prevent users from deleting or adding members to a personal workspace
-  Add the ability to add users to a workspace (if not personal)
-  User can leave workspace
- Fetch user data me query
- Mutation to change the username
- Vite environment variables config

## Pages added
- Homepage (dashboard) which checks if the user logged in or not, and shows a welcome message
- Account page where you can change the user's name
- Workspace list page where you can view a list of workspace, and add a new one
- Workspace details page where you can 
  - edit/delete workspace
  - add/edit/delete members
  - Change members role

## Notes
Folder structure and significant part of code inspired by Re:Earth web project.   
Some code parts will be divided further when layout (design task) and modals are added.
This code passed the acceptance test.


web/src/state/index.ts
```diff
@@ -3,3 +3,14 @@ import { atom, useAtom } from "jotai";
...
+
+export type Wrokspace = {
+  id: string;
+  name: string;
+  members?: Array<any>;
+  assets?: any;
+  projects?: any;
+  personal?: boolean;
+};
+const workspace = atom<Wrokspace | undefined>(undefined);
```
KaWaite, line 15:
```suggestion
const workspace = atom<Workspace | undefined>(undefined);
```

web/src/gql/queries/workspace.ts
```diff
@@ -0,0 +1,156 @@
...
+        name
+        email
+      }
+      userId
+      role
+    }
+    personal
+  }
+
+  query workspaces {
```
KaWaite, line 19: query GetWorkspaces

web/src/gql/queries/workspace.ts
```diff
@@ -0,0 +1,156 @@
...
+      user {
+        id
+        name
+        email
+      }
+      userId
+      role
+    }
+    personal
+  }
```
KaWaite, line 17: I'm pretty sure you can pull this out as its own variable and use it inside UPDATE_WORKSPACE and ADD_MEMBER_TO_WORKSPACE as well!
If you want you can also create a fragments folder and put it there (like we do in Re:Earth)? But at least move it out since you use this structure in update and add members!

web/src/gql/queries/workspace.ts
```diff
@@ -0,0 +1,156 @@
+import { gql } from "@apollo/client";
+
+export const WORKSPACES = gql`
```
KaWaite, line 3: GET_WORKSPACES

web/src/gql/queries/workspace.ts
```diff
@@ -0,0 +1,156 @@
...
+      workspaces {
+        id
+        ...Workspace
+      }
+    }
+  }
+`;
+
+export const UPDATE_WORKSPACE = gql`
+  mutation updateWorkspace($workspaceId: ID!, $name: String!) {
```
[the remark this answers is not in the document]
KaWaite: Please update all where necessary

web/src/gql/queries/user.ts
```diff
@@ -0,0 +1,116 @@
+import { gql } from "@apollo/client";
+
+export const SEARCH_USER = gql`
+  query searchUser($nameOrEmail: String!) {
+    searchUser(nameOrEmail: $nameOrEmail) {
```
KaWaite, line 5: Keeping it same as Re:Earth
```suggestion
export const GET_USER_BY_SEARCH = gql`
  query GetUserBySearch($nameOrEmail: String!) {
    searchUser(nameOrEmail: $nameOrEmail) {
```

web/src/gql/queries/user.ts
```diff
@@ -0,0 +1,116 @@
...
+      id
+      name
+      email
+    }
+  }
+`;
+
+export const ME = gql`
+  query Me {
```
KaWaite, line 14: Queries should start with GetXxxx

```suggestion
export const GET_ME = gql`
  query GetMe {
```

web/src/gql/provider.tsx
```diff
@@ -42,11 +43,15 @@ const Provider: React.FC<Props> = ({ children }) => {
     }
   });
 
+  const httpLink = new HttpLink({
+    uri: endpoint,
+  });
```
KaWaite, line 48: I don't think this is necessary. The Apollo docs [here](https://www.apollographql.com/docs/react/api/link/apollo-link-http/) say Apollo uses Httplink by default if we pass the uri option to ApolloClient, which we do.

web/src/App.tsx
```diff
@@ -1,18 +1,25 @@
...
 import DashboardPage from "@reearth-cms/components/pages/Dashboard";
 import RootPage from "@reearth-cms/components/pages/RootPage";
+import WorkspacePage from "@reearth-cms/components/pages/Workspace";
 import { Provider as GqlProvider } from "@reearth-cms/gql";
 import { Provider as I18nProvider } from "@reearth-cms/i18n";
 import { BrowserRouter as Router, useRoutes } from "react-router-dom";
 
 import "./App.css";
 import "antd/dist/antd.css";
+import WorkspaceListPage from "./components/pages/WorkspaceList";
```
KaWaite, line 13: Can you move this up where it seems more appropriate?

web/src/App.tsx
```diff
@@ -1,18 +1,25 @@
...
 
 function AppRoutes() {
   const routes = useRoutes([
     { path: "/", element: <RootPage /> },
     { path: "/dashboard", element: <DashboardPage /> },
+    { path: "/dashboard/:workspaceId", element: <DashboardPage /> },
+    { path: "/workspace/:workspaceId", element: <WorkspacePage /> },
+    { path: "/workspace", element: <WorkspaceListPage /> },
```
KaWaite, line 21: Can you rename the path to `/workspaces`. I feel like it makes more sense that /workspaces shows the list, and then /workspaces/:workspaceId would should a workspace from the workspace list.
If you disagree please debate :P

web/src/App.tsx
KaWaite: We should actually setup import filtering like we do in Re:Earth so we don't have to worry about the order everytime

web/src/components/organisms/Settings/Workspace/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+import Form from "@reearth-cms/components/atoms/Form";
+import Input from "@reearth-cms/components/atoms/Input";
+import List from "@reearth-cms/components/atoms/List";
+import React, { useCallback, useEffect, useState } from "react";
+import { useNavigate, useParams } from "react-router-dom";
+
+import useHooks, { RoleUnion } from "./hooks";
+
+const Workspace: React.FC = () => {
+  const { workspaceId } = useParams();
```
KaWaite, line 12: Is there any reason you do `useParams` here instead of passing the workspaceId down from the Workspace Page as a prop?

web/src/components/organisms/Settings/Workspace/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+    updateName,
+    addMembersToWorkspace,
+    updateMemberOfWorkspace,
+    removeMemberFromWorkspace,
+  } = useHooks({ workspaceId });
+  const [owner, setOwner] = useState(false);
+  const [memberName, setMemberName] = useState<string>("");
+  const [workspaceName, setWorkspaceName] = useState<string>("");
+  const navigate = useNavigate();
+  const members = currentWorkspace?.members;
```
KaWaite, line 29: The top of components and hooks often gets messy. Can you arrange these and space them out a little more cleanly? (I typically group third party hooks, useHooks, and stateful things separately. But its up to you since it depends on the comp/hooks file)

web/src/components/organisms/Settings/Workspace/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+  const handleMemberNameChange = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      setMemberName?.(e.currentTarget.value);
+      searchUser(e.currentTarget.value);
```
KaWaite, line 34: Can you explain why you need to search the user after a name change?

web/src/components/organisms/Settings/Workspace/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+    },
+    [setMemberName, searchUser]
+  );
+
+  const handleAddMember = useCallback(() => {
```
KaWaite, line 39: Please make sure the function naming is in the `handleNounVerb` format. If we do that, it cleans everything up and makes it easier to read.

handleMemberAdd
handleMemberRemove
handleWorkspaceAdd
handleWorkspaceUpdate

etc

web/src/components/organisms/Settings/Workspace/hooks.ts
```diff
@@ -0,0 +1,224 @@
...
+    currentWorkspace,
+    searchedUser,
+    changeSearchedUser,
+    createWorkspace,
+    updateName,
+    deleteWorkspace,
+    searchUser,
+    addMembersToWorkspace,
+    updateMemberOfWorkspace,
+    removeMemberFromWorkspace,
```
KaWaite, line 220: handleXxxxx format

web/src/components/organisms/Settings/Workspace/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+    await deleteWorkspace();
+    navigate("/workspace");
+  }, [deleteWorkspace, navigate]);
+
+  const handleDeleteMember = useCallback(
+    async (userId: string) => {
+      await removeMemberFromWorkspace(userId);
+    },
+    [removeMemberFromWorkspace]
+  );
```
KaWaite, line 56: Is there a reason you make separate callbacks in the hooks.ts file, and then wrap them again in callbacks here in the component?
It feels redundant and the code for both files are very long, reducing the benefits of a hooks file.

web/src/components/organisms/Dashboard/hooks.ts
```diff
@@ -0,0 +1,87 @@
...
+      }
+      refetch();
+    },
+    [createWorkspaceMutation, setCurrentWorkspace, refetch, navigate]
+  );
+
+  return {
+    user,
+    workspaces,
+    currentWorkspace: workspace as Workspace,
```
KaWaite, line 83: instead of doing `as Workspace` here, can you not type the useState as `Workspace` or `Workspace | undefined` to avoid type casting?

web/src/components/organisms/Settings/workspaceList/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+type Props = {
+  workspaceId: string;
+};
+
+const WorkspaceList: React.FC<Props> = ({ workspaceId }) => {
+  const { workspaces, createWorkspace, selectWorkspace } = useHooks({
+    workspaceId,
+  });
+  const navigate = useNavigate();
+  const [workspaceName, setWorkspaceName] = useState<string>("");
```
KaWaite, line 21: For useState that are strings, you do both `useState("")` and `useState<string>("")`. Can we stick to one (preferably the first one, since the type can be inferred by the "")??

web/src/components/pages/Workspace/index.tsx
```diff
@@ -0,0 +1,16 @@
+import { AuthenticationRequiredPage } from "@reearth-cms/auth";
+import Workspace from "@reearth-cms/components/organisms/Settings/Workspace";
+
+export type Props = {
+  path?: string;
+};
+
```
KaWaite, line 7: Aren't we getting the workspaceId here from the url?

web/src/components/pages/WorkspaceList/index.tsx
```diff
@@ -0,0 +1,15 @@
+import { AuthenticationRequiredPage } from "@reearth-cms/auth";
+import WorkspaceList from "@reearth-cms/components/organisms/Settings/workspaceList";
+
+export type Props = {
+  path?: string;
+  workspaceId?: string;
```
KaWaite, line 6: I think you have this code mixed up between this Page and the Workspace Page

KaWaite (commented):
Please review my comments. Some of them are asking for changes in more than just the place where the comment is (in particular, about naming)!

web/src/components/organisms/Settings/Workspace/index.tsx
issmail-basel: Will check this later when I start with layout, to check if I need to keep the current workspace param and pass it to children or capsulate the param as I did here.

web/src/components/organisms/Settings/Workspace/index.tsx
issmail-basel: To get user suggestions while typing the same as `re:earth` member modal (when adding a new member to the workspace you have to search for him first)

web/src/config.ts
```diff
@@ -10,11 +10,15 @@ declare global {
...
-  api: "/api",
-  auth0Audience: "https://api.test.reearth.dev",
-  auth0Domain: "reearth-oss-test.eu.auth0.com",
-  auth0ClientId: "k6F1sgFikzVkkcW9Cpz7Ztvwq5cBRXlv",
+  api: env.REEARTH_CMS_API || "/api",
+  auth0Audience:
+    env.REEARTH_CMS_AUTH0_AUDIENCE || "https://api.test.reearth.dev",
+  auth0Domain: env.REEARTH_CMS_AUTH0_DOMAIN || "reearth-oss-test.eu.auth0.com",
+  auth0ClientId:
+    env.REEARTH_CMS_AUTH0_CLIENT_ID || "k6F1sgFikzVkkcW9Cpz7Ztvwq5cBRXlv",
```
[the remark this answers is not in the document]
issmail-basel: There is already a default `client ID` in Re:earth to "**start app with zero configuration**"
and also you can access it with `window.REEARTH_WEB_AUTH0_CLIENT_ID` or check the config file from the network tab.
So it's not sensitive on the front-side

web/src/gql/provider.tsx
issmail-basel: When using `link: ApolloLink.from` option
It does not work, unless I explicitly provide the HttpLink, for now, I'll leave it, and maybe in the future I'll inspect why this happening and check graphQL repo for similar issues

web/src/components/organisms/Dashboard/hooks.ts
```diff
@@ -0,0 +1,87 @@
...
+  useEffect(() => {
+    if (workspace?.id && workspace.id !== currentWorkspace?.id) {
+      setCurrentWorkspace({
+        personal,
+        ...workspace,
+      });
+    }
+  }, [currentWorkspace, workspace, setCurrentWorkspace, personal]);
+
+  const changeWorkspace = useCallback(
```
KaWaite, line 51: Here should also be updated to be handleXXXX. Also `createWorkspace` below

web/src/components/pages/WorkspaceList/index.tsx
KaWaite: Sorry I should have explained what I meant better. This is the WorkspaceList page, which doesn't receive workspaceId from the path. Whereas the Workspace page DOES receive the workspaceId from the path. Shouldn't these props be in the Workspace component, and this be empty?

web/src/components/organisms/Settings/Workspace/hooks.ts
```diff
@@ -0,0 +1,225 @@
...
+import { useCallback, useEffect, useState } from "react";
+import { useNavigate } from "react-router-dom";
+
+export type RoleUnion = "READER" | "WRITER" | "OWNER";
+
+type Params = {
+  workspaceId: string | undefined;
+};
+
+export default (params: Params) => {
```
KaWaite, line 23: Can you destructure your props and for consistency can you rename the type to `Props`?

web/src/components/organisms/Settings/Workspace/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  const handleMemberAdd = useCallback(() => {
+    if (!searchedUser) return;
+    handleMemberAddToWorkspace([searchedUser.id]);
+    setMemberName("");
+    changeSearchedUser(undefined);
+  }, [searchedUser, handleMemberAddToWorkspace, changeSearchedUser]);
+
+  const handleWorkspaceDeleteAction = useCallback(async () => {
```
KaWaite, line 49: Sorry for all the small asks, but can you do 
```suggestion
  const handleWorkspaceDeletion = useCallback(async () => {
```

web/src/components/organisms/Settings/Account/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const handleInputChange = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      setFilterText?.(e.currentTarget.value);
+    },
+    [setFilterText]
+  );
+
+  const handleClick = useCallback(() => {
+    updateName(filterText);
+  }, [filterText, updateName]);
```
KaWaite, line 24: Since you already have a hooks.ts file, can you go the whole nine yards and move all this into there as well?

KaWaite (approved):
Ok, I left a couple more comments if you can update those I think its okay to merge.

PR #67, opened 2022-05-15
ci: add workflow for deploying and releasing

KeisukeYamashita:
## WHY

Like `reearth-backend`, build and release the image to the DockerHub and the GCP project for test env for CI/CD.


HideBa (approved):
I had a review and pushed small fixes. 
I'll take over the rest of the tasks to make infra completed. 
Thanks!

PR #84, opened 2022-06-08
feat(web): user and workspace

issmail-basel:


web/src/App.tsx
```diff
@@ -11,12 +11,14 @@ import { BrowserRouter as Router, useRoutes } from "react-router-dom";
...
 import "./App.css";
 import "antd/dist/antd.css";
+import MembersPage from "./components/pages/Members";
 
 function AppRoutes() {
   const routes = useRoutes([
     { path: "/", element: <RootPage /> },
     { path: "/dashboard", element: <DashboardPage /> },
     { path: "/dashboard/:workspaceId", element: <DashboardPage /> },
+    { path: "/members/:workspaceId", element: <MembersPage /> },
```
KaWaite, line 21: Up for debate, but logically speaking wouldn't it be better if this was `/workspaces/:workspaceId/members`?
For me it makes more sense because `/members` is probably never going to make sense for us. If we were going to have `/members/id/projects` or something, okay, but we are workspace based and the members are related to workspaces.

web/src/components/organisms/Settings/Members/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+import MemberRoleModal from "@reearth-cms/components/molecules/Member/MemberRoleModal";
+import { PageHeader, Table, Modal } from "antd";
+import Avatar from "antd/lib/avatar/avatar";
+import Search from "antd/lib/input/Search";
+import Layout, { Header, Content } from "antd/lib/layout/layout";
+import Sider from "antd/lib/layout/Sider";
+import React, { useCallback, useState } from "react";
+import { useParams } from "react-router-dom";
+
+import useDashboardHooks from "../../Dashboard/hooks";
```
KaWaite, line 20: If these hooks will be used in multiple places, can we think about having either a Common folder like in ReEarth where we have a common hooks file, or at least in the atomic component directory have something like `organisms/commonHooks.ts`?
It would allow us to consolidate repeating hooks. If that file got too big we could consider splitting into categorical hooks. 
What do you think?

web/src/components/organisms/Settings/Members/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+    if (!searchedUser) return;
+    handleMemberAddToWorkspace([searchedUser.id]);
+    changeSearchedUser(undefined);
+    handleMemberCreationModalClose();
+  }, [
+    searchedUser,
+    handleMemberAddToWorkspace,
+    changeSearchedUser,
+    handleMemberCreationModalClose,
+  ]);
```
KaWaite, line 155: Can any of this be 1) Moved to the molecule components? Or 2) Be moved into the hooks file?
Especially functions like
```
const handleRoleChange = useCallback(
    (member: any) => {
      handleRoleModalOpen(member);
    },
    [handleRoleModalOpen]
  );
```
I feel like they aren't needed. You aren't doing anything new that the original `handleRoleModalOpen` isn't doing.

web/src/components/organisms/Settings/Members/hooks.ts
```diff
@@ -0,0 +1,186 @@
...
+        return;
+      }
+      setWorkspace(workspace);
+      // TODO: notification
+    },
+    [workspaceId, removeMemberFromWorkspaceMutation, setWorkspace]
+  );
+
+  const handleRoleModalClose = useCallback(
+    (r?: boolean) => {
```
KaWaite, line 135: What is the `r` for? Is it for role, or refetch? Can you name it a little more meaningfully?

web/src/components/organisms/Dashboard/index.tsx
```diff
@@ -1,28 +1,128 @@
...
 export type Props = {
   workspaceId?: string;
 };
 
-const Dashboard: React.FC<Props> = ({ workspaceId }) => {
-  const { logout } = useAuth();
-  const navigate = useNavigate();
-
-  const { user } = useHooks(workspaceId);
+const Dashboard: React.FC<Props> = () => {
```
KaWaite, line 19: I don't think you need Props here. You aren't passing the workspaceId, you are getting it from useParams, which takes it from the url, right?

web/src/components/organisms/Dashboard/index.tsx
```diff
@@ -1,28 +1,128 @@
...
+            currentWorkspace={currentWorkspace}
+            user={user}
+          ></MoleculeHeader>
+        </Header>
+        <Layout>
+          <Sider
+            collapsible
+            collapsed={collapsed}
+            onCollapse={(value) => setCollapsed(value)}
+            style={{ backgroundColor: "#fff" }}
```
KaWaite, line 50: For CMS should we use inline styles? Or, even if its just one style, should we use styled components? I feel like we should use styled components to keep all custom styles in the same place.

web/src/components/molecules/Common/Header/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+      navigate(`/dashboard/${id}`);
+    },
+    [navigate]
+  );
+
+  const WorkspacesMenu = (
+    <HeaderMenu
+      items={[
+        {
+          label: "Presonal Account",
```
KaWaite, line 53:
```suggestion
          label: "Personal Account",
```

web/src/components/molecules/Common/WorkspaceCreationModal/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+      })
+      .catch((info) => {
+        console.log("Validate Failed:", info);
+      });
+  }, [form, onClose, onSubmit]);
+
+  const handleClose = useCallback(() => {
+    onClose?.(true);
+  }, [onClose]);
+  return (
```
KaWaite, line 41:
```suggestion
  }, [onClose]);
  
  return (
```

web/src/components/molecules/Common/WorkspaceCreationModal/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+        form.resetFields();
+      })
+      .catch((info) => {
+        console.log("Validate Failed:", info);
+      });
+  }, [form, onClose, onSubmit]);
+
+  const handleClose = useCallback(() => {
+    onClose?.(true);
+  }, [onClose]);
```
KaWaite, line 40: Same as another comment. Is this necessary? I am assuming you are trying to avoid unecessary re-renders, but if onClose is wrapped in a useCallback, isn't that enough?

web/src/components/molecules/Common/WorkspaceCreationModal/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return (
+    <Modal visible={open} onCancel={handleClose} onOk={handleSubmit}>
+      {/* {formik.isSubmitting} */}
```
KaWaite, line 43: Can you delete

web/src/components/molecules/Common/WorkspaceCreationModal/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <Form form={form} layout="vertical" initialValues={initialValues}>
+        <Form.Item
+          name="name"
+          label="Workspace name"
+          rules={[
+            { required: true, message: "Please input the title of workspace!" },
```
KaWaite, line 49:
```suggestion
            { required: true, message: "Please input the title of the current workspace!" },
```

web/src/components/molecules/Common/WorkspaceMenu/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+  icon: any,
+  show: "personal" | "notPersonal" | "both"
+) {
+  return {
+    key,
+    icon,
+    label,
+    show,
+  };
+}
```
KaWaite, line 33: This function doesn't really do anything. You can keep things simpler by just write out the objects themselves.

web/src/components/molecules/Member/MemberRoleModal/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+        layout="vertical"
+        initialValues={{
+          userId: member?.userId,
+          role: member?.role,
+        }}
+      >
+        <Form.Item
+          name="role"
+          label="Role"
+          rules={[{ required: true, message: "Please input the role!" }]}
```
KaWaite, line 69:
```suggestion
          rules={[{ required: true, message: "Please input the appropriate role for this member!" }]}
```

web/src/index.css
```diff
@@ -1,13 +1,35 @@
...
+
+.ant-layout-sider-children {
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+}
+
+.ant-dropdown-menu-item-group-title {
+    color: #8C8C8C!important;
+}
```
KaWaite, line 35: I am guessing you are using these in specific places. Any way you can be more specific so you can avoid using `!important`?

KaWaite (commented):
Reviewed!

web/src/components/molecules/Common/WorkspaceMenu/index.tsx
```diff
@@ -43,13 +33,43 @@ const WorkspaceMenu: React.FC<Props> = ({
...
-    getItem("Integration", "integration", <ApiOutlined />, "both"),
-    getItem("Role", "role", <UserSwitchOutlined />, "notPersonal"),
-    getItem("API key", "api-key", <SearchOutlined />, "both"),
-    getItem("Settings", "settings", <SettingOutlined />, "notPersonal"),
+  const items: WorkspaceItemType[] = [
+    {
+      label: "Member",
+      key: "member",
+      icon: <UsergroupAddOutlined />,
+      show: "notPersonal" as MenuShowType,
```
KaWaite, line 41: Is the `as MenuShowType` actually necessary to avoid type errors? I feel like as long as the value is one of the three string values in MenuShowType it shouldn't cause an error, but maybe im wrong?

web/src/components/molecules/Common/WorkspaceMenu/index.tsx
```diff
@@ -59,7 +79,7 @@ const WorkspaceMenu: React.FC<Props> = ({
 
   const onClick = (e: any) => {
```
KaWaite, line 80: Please do a if/else if instead of just if/if. That way if the first if passes, it knows it doesn't need to check the next one.

web/src/components/molecules/Common/WorkspaceMenu/index.tsx
```diff
@@ -18,22 +18,12 @@ export interface Props {
   defaultSelectedKeys?: string[];
 }
```
KaWaite, line 20: Any reason you are importing directly from antd for the Menu component and not creating an atom like the other antd components?

web/src/components/molecules/Dashboard/Greeting.tsx
```diff
@@ -0,0 +1,31 @@
+import styled from "@emotion/styled";
+import React from "react";
+
+const Greeting: React.FC = () => {
+  return <DashboardCard>Welcome to Re:Earth CMS !</DashboardCard>;
```
KaWaite, line 5: Are we not doing internationalization yet? I know it is setup, so a little curious why we are avoiding it at the moment.

web/src/components/molecules/Member/MemberCreationModal/index.tsx
```diff
@@ -72,19 +72,21 @@ const MemberCreationModal: React.FC<Props> = ({
     form
       .validateFields()
       .then(async () => {
-        if (searchedUser?.id) await onSubmit?.();
+        if (searchedUser?.id) await onSubmit?.([searchedUser.id]);
+        onClose?.(true);
         form.resetFields();
```
KaWaite, line 77: Does this modal actually create a member, or add a member to a workspace? If it is the latter, the name for this component is confusing and maybe `MemberAddModal` is more clear.

web/src/components/organisms/Settings/Members/hooks.ts
```diff
@@ -131,32 +134,20 @@ export default ({ workspaceId }: Props) => {
...
-      }
-    },
-    [refetch]
-  );
+  const handleRoleModalClose = useCallback(() => {
+    setRoleModalShown(false);
+    setSelectedMember(undefined);
+  }, []);
 
   const handleRoleModalOpen = useCallback((member: any) => {
```
KaWaite, line 142: `member: Member`

web/src/components/organisms/Settings/Members/index.tsx
```diff
@@ -85,7 +68,7 @@ const Members: React.FC = () => {
 
   const members = currentWorkspace?.members;
 
-  const showConfirm = useCallback(
+  const handleMemberDelete = useCallback(
```
KaWaite, line 71: This callback still is accepting anything as member. Now that we have the Member type we can type it, right?

web/src/components/organisms/Project/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+    handleWorkspaceCreate,
+    handleModalClose,
+    handleModalOpen,
+    modalShown,
+  } = useDashboardHooks(workspaceId);
+
+  return (
+    <>
+      <PaddedContent>
+        <Greeting></Greeting>
```
KaWaite, line 25: `<Greeting />

web/src/components/organisms/Project/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+              Search
+            </Button>
+          </ButtonWrapper>
+        </ActionHeader>
+      </PaddedContent>
+      <WorkspaceCreationModal
+        open={modalShown}
+        onClose={handleModalClose}
+        onSubmit={handleWorkspaceCreate}
+      ></WorkspaceCreationModal>
```
KaWaite, line 44:
```
  <WorkspaceCreationModal
        open={modalShown}
        onClose={handleModalClose}
        onSubmit={handleWorkspaceCreate}
      />
```

KaWaite (commented):
So a couple questions:
- Is there a reason we have stopped indirectly using antd components (aka importing them as atoms, and then using those atoms)?
- Is there a reason we aren't setting up the text to be using i18next? I know it is setup, and seems like it'll be really difficult later to go back through all the text and wrap it in the internationalization.
- Did you remove workspaces from this PR?

web/src/components/molecules/Common/WorkspaceMenu/index.tsx
issmail-basel: Typescript is showing type errors, I couldn't solve it only by casting like this.
I'll check this later as well because I'll get back to this menu and it will be linked to other pages.

web/src/components/molecules/Common/WorkspaceMenu/index.tsx
issmail-basel: I left it like this for the future, as there will be more menu items.
I'll make it if/else and change it later to something more suitable

web/src/components/molecules/Dashboard/Greeting.tsx
issmail-basel: I worked on translations on a separate branch with the account page (where user can change their language preferences).
So after this PR will add that one

KaWaite (approved):
Looks good! T
he only thing is there are several places where you write `<Component></Component>` without any children. Whether before merging this one, or in a refactor, please update them to `<Component />`!

PR #109, opened 2022-07-19
feat(server): add task runner interface

HideBa:
# Why this interface
This asynchronous processing will use an architecture that allows the Publisher to know and control the Subscriber's presence, rather than a PubSub-like format that achieves loose coupling. This is, in other words, a queuing architecture. Therefore, as an abstraction, we have decided to represent the queue as a group of processes of a certain purpose and each concrete request as a task.

# Description of each
* Queue: A queue is a processing system that is responsible for a certain process. In this case, one queue is the process of decompressing compressed files. If other asynchronous processes such as sending e-mails are added in the future, they will be created in separate queues.
* Task: Task refers to each request that is added to the queue, including the Payload that should be communicated to the Subscriber. In this case, the Payload includes a link to the uploaded Zip file, etc.

# Concern
There is currently no way to guarantee the data type of data exchanged between Publisher and Subscriber. Cloud Tasks, which will be used this time, seems to have an SDK that uses proto for typing, and it is possible to use such SDK. However, there is currently no other asynchronous processing, and we have decided that it is not worth the cost to introduce them for this task, so we will not adopt them this time.

# Point of Reviewing
* As a way of abstraction, does the queue, task approach seem like a good way to go?
* Omissions in the implementation of minor details are not required to be reviewed this time.


server/internal/usecase/gateway/queue.go
```diff
@@ -0,0 +1,16 @@
...
+type CompressFileExtention string
+
+const (
+	Zip = CompressFileExtention("zip")
+)
+
+// *********************************
+type Queue interface {
+	AddTask(ctx context.Context) (queueId string, err error)
+}
```
rot1024, line 16: There are several points to consider.

- It is not good if the task is not described in any detail at all at gateway. Whether it is GCP or AWS, the request to unpack a Zip itself is not the concern of the infrastructure layer, it is unquestionably the domain. The infrastructure layer is interested in how to perform the task it is handed, not what it is. Therefore, the details of the task should be describable at the domain or use case layer.
- In the description of the decompression task, it is not necessary to specify the decompression format explicitly. The actual filename of the saved asset will tell you, and if it is in a format that cannot be decompressed, it will return an error.
- You cannot get the results of a task directly. If you want the results, you will receive them indirectly by having an outside party notify the API exposed by the CMS.
- The word queue has been eliminated from the interface name. Whether or not to queue a task is a technical concern. It is even possible to execute tasks without queuing. For example, a task runner implementation could be used for unit testing, which is completed synchronously.

```go
// domain model (pkg/task)

type Payload struct {
	ExtractAsset  *ExtractAssetPayload
	CompressAsset *CompressAssetPayload
}

type ExtractAssetPayload struct {
	Asset id.AssetID
}

func (t *ExtractAssetPayload) Payload() Payload {
	return Payload {
		ExtractAsset: t
	}
}

type CompressAssetPayload struct {
	Asset id.AssetID
}

func (t *CompressAssetPayload) Payload() Payload {
	return Payload {
		CompressAsset: t
	}
}

type CompressZipPayload struct {
	Asset id.AssetID
}

type ID = id.TaskID

// gateway

type TaskRunner interface {
	Run(context.Context, Payload) (id.TaskID, error)
}
```

server/internal/infrastructure/gcp/cloudtasks/cloudtasks.go
```diff
@@ -0,0 +1,45 @@
+package cloudtasks
```
rot1024, line 1: Do not create more folders than you need. Keep the folder hierarchy as shallow as possible.

`gcp/cloudtask/cloudtask.go` -> `gcp/cloudtask.go`

Since this is just queuing the passed tasks, the amount of implementation should not be that large.

rot1024 (changes requested):
- Move drawio svg to `docs` directory ([refer](https://github.com/golang-standards/project-layout/tree/master/docs))
- Add `(server)` scope to pr title

server/internal/usecase/gateway/queue.go
HideBa: I found that it is not a technical issue whether it is asynchronous or not, but that the process of decompression is placed as business logic. We have reflected the review and ask that you review it again. 😄 

I fixed where you pointed out with following commit [6ae4627](https://github.com/reearth/reearth-cms/pull/109/commits/6ae46276c4c8105f5906c301a3e53e96b0172c12)

HideBa:
@rot1024 I've moved the diagram under docs folder with [70f9a44](https://github.com/reearth/reearth-cms/pull/109/commits/70f9a445ff4c3cfddc8bafbf7147eac0e13cff80).

rot1024 (approved):

PR #112, opened 2022-07-20
feat(server): init worker app with multi-module workspaces

HideBa:
# What
* Create worker dir
* Introduce [multi-module workspaces](https://go.dev/doc/tutorial/workspaces) so that we can have multiple modules even in the same workspace 
* Add Dockerfile
* Add a simple server which returns "hello world"

# Why
* Since Go1.18 supports the multi-module workspaces feature which allows us to have multiple modules under a single workspace though we couldn't do that and were forced to have a hacky way, I've introduced that to take the full advantage of mono-repo. 

# Point of Review
* If there is any problem with directory structure, Dockerfile, etc

# Things you don't need to review
* The implementation of a simple server since it's a temporary development
* GitHub actions workflow file will be implemented later

### DIFF
--- a/worker/Dockerfile
+++ b/worker/Dockerfile
@@ -12,12 +12,12 @@ RUN go mod download
 COPY cmd/ /app/cmd/
 COPY internal/ /app/internal/
 
-RUN CGO_ENABLED=0 go build -tags "${TAG}" "-ldflags=-X main.version=${VERSION} -s -w -buildid=" -trimpath ./cmd/reearth-worker
+RUN CGO_ENABLED=0 go build -tags "${TAG}" "-ldflags=-X main.version=${VERSION} -s -w -buildid=" -trimpath ./cmd/reearth-cms-worker
 
 FROM scratch
 
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
-COPY --from=build /app/reearth-worker /app/reearth-worker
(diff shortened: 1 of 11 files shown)


worker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+COPY cmd/ /app/cmd/
+COPY internal/ /app/internal/
+
+RUN CGO_ENABLED=0 go build -tags "${TAG}" "-ldflags=-X main.version=${VERSION} -s -w -buildid=" -trimpath ./cmd/reearth-worker
+
+FROM scratch
+
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --from=build /app/reearth-worker /app/reearth-worker
```
rot1024, line 20:
```suggestion
RUN CGO_ENABLED=0 go build -tags "${TAG}" "-ldflags=-X main.version=${VERSION} -s -w -buildid=" -trimpath ./cmd/reearth-cms-worker

FROM scratch

COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
COPY --from=build /app/reearth-cms-worker /app/reearth-cms-worker
```

rot1024 (changes requested):
Yes it's time to use go workspace!

rot1024 (approved):

PR #123, opened 2022-07-22
feat(web): asset ui atom components

nourbalaha:
In this PR I did the following:
1- Added AntDesign's ProTable, Radio, Tree and Upload components as atom components.
2- Added DownloadButton as atom component to download any file by providing a file name and a url.
3- Added TilesetPreview as atom component to preview tileset files.
4- Added CustomTag as atom component.
5- updated `vite.config.ts` file to make Cesium and ProTable work.

### DIFF
--- a/web/src/components/atoms/CustomTag/index.tsx
+++ b/web/src/components/atoms/CustomTag/index.tsx
@@ -7,11 +7,9 @@ type CustomTagProps = {
 };
 
 const CustomTag: React.FC<CustomTagProps> = ({ value, color }) => {
-  const defaultColor = "#87e8de";
-  const defaultValue = "";
   return (
-    <CustomTagWrapper color={color ?? defaultColor}>
-      <span>{value ?? defaultValue}</span>
+    <CustomTagWrapper color={color ?? "#87e8de"}>
+      <span>{value ?? ""}</span>
     </CustomTagWrapper>
   );
 };
--- a/web/src/components/atoms/CustomTag/index.tsx
+++ b/web/src/components/atoms/CustomTag/index.tsx
@@ -1,0 +1,36 @@
+import styled from "@emotion/styled";
+import { CSSProperties } from "react";
+
+type CustomTagProps = {
+  value?: number | string;
+  color?: CSSProperties["color"];
+};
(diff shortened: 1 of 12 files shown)


web/src/components/atoms/DownloadButton/index.tsx
```diff
@@ -0,0 +1,37 @@
+import { DownloadOutlined } from "@ant-design/icons";
```
KaWaite, line 1: Download this into the Icon atom and use from there.

web/src/components/atoms/DownloadButton/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+import Button from "@reearth-cms/components/atoms/Button";
+import { ButtonProps } from "antd/lib/button/button";
```
KaWaite, line 3: Add this to the Button atom and import with the Button

web/src/components/atoms/CustomTag/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { CSSProperties } from "react";
+
+type CustomTagProps = {
+  value?: number | string;
+  color?: CSSProperties["color"];
+};
+
+const CustomTag: React.FC<CustomTagProps> = ({ value, color }) => {
+  const defaultColor = "#87e8de";
+  const defaultValue = "";
```
KaWaite, line 11: Either just use these values directly where they are used, or move them outside of the component as they won't ever change

KaWaite (commented):
Please see comments!

web/src/components/atoms/Icon/index.tsx
```diff
@@ -1,5 +1,6 @@
-import { HomeOutlined } from "@ant-design/icons";
+import { HomeOutlined, DownloadOutlined } from "@ant-design/icons";
 
 export default {
   home: HomeOutlined,
+  download: DownloadOutlined,
```
KaWaite, line 5: Thanks for changing this. But sorry I think this MIGHT change in another issue sometime later. @Basel-Issmail and I are chatting about it.

PR #139, opened 2022-07-27
feat(web): add antd icons as atom

issmail-basel:


web/src/components/atoms/Icon/index.tsx
```diff
@@ -1,6 +1,91 @@
-import { HomeOutlined, DownloadOutlined } from "@ant-design/icons";
+import styled from "@emotion/styled";
```
KaWaite, line 1: Is there any reason at the moment you made this component to be useable with more than just the antd icons?
Do we have icons we are using that aren't antd that we want to support? Or are you just keeping this as close or the same to Re:Earth as possible?

web/src/components/atoms/Icon/index.tsx
```diff
@@ -1,6 +1,91 @@
...
+        color={color}
+        style={style}
+        alt={alt}
+        size={sizeStr}
+        onClick={onClick}
+      />
+    );
+  }
+
+  return React.createElement(src); // this is of type (React.ForwardRefExoticComponent)
```
KaWaite, line 63: Does this still allow for antd Icon's props? (Like spin, rotate, etc)?
Or does @lavalse design not really make use of those at all?

issmail-basel: @KaWaite , for now, there is no use for antd properties, but can you help me with code, and find a way to use and props with it?

issmail-basel: @KaWaite , can be added when needed, for now, there is no usage for these props.
For now, there is no need (to keep components as generic as possible and to have a way to replace antd icons)
in the future if we need any prop we can simply:
`React.createElement(src, {propName: propValue})`

web/src/components/atoms/Icon/index.tsx
issmail-basel: @KaWaite , yes, to have a generic component, where we can use multiple icons formats (to keep components as generic as possible and to have a way to replace antd icons)

KaWaite (approved):

PR #152, opened 2022-08-01
feat(web): asset ui organisms and pages

nourbalaha:
In this PR I added the following:
1- AssetList's organisms and pages.
2- Asset's organisms and pages.
3- required utils for AssetList's organisms and pages.
4- implementations of list assets, upload asset and update asset.
5- updated asset.type according to the new schema.

Still work in progress (not in this PR):
1- asset list table pagination.
2- asset list table sort.
3- asset list table search.
4- upload modal url and google drive tabs.

Please use the backend of [feat/assets-usecase-adapter](https://github.com/reearth/reearth-cms/tree/feat/assets-usecases-adapter) branch to test.


web/src/gql/provider.tsx
```diff
@@ -44,7 +49,7 @@ const Provider: React.FC<Props> = ({ children }) => {
 
   const client = new ApolloClient({
     uri: endpoint,
-    link: ApolloLink.from([errorLink, authLink, httpLink]),
+    link: ApolloLink.from([errorLink, authLink, uploadLink, httpLink]),
```
KaWaite, line 52: (Just learnt this recently) Apparently you don't need httpLink if you have uploadLink.
So please uninstall httpLink and remove from here.

Also, after you do that, if you have an issue with the createUploadLink type being wrong, you'll have to downgrade the @type file to 17.0.0.
If you have no issues, please let me know, because maybe apollo-client's newest version fixed a type issue.

web/src/components/organisms/Asset/Asset/hooks.ts
```diff
@@ -0,0 +1,81 @@
...
+import { Asset } from "@reearth-cms/components/molecules/Asset/asset.type";
+import { viewerRef } from "@reearth-cms/components/molecules/Asset/Asset/AssetBody/index";
+import {
+  PreviewType,
+  useGetAssetQuery,
+  useUpdateAssetMutation,
+} from "@reearth-cms/gql/graphql-client-api";
+
+export default (assetId?: string) => {
+  const [asset, setAsset] = useState<Asset>({} as Asset);
```
KaWaite, line 12: If this works I think its better
```suggestion
  const [asset, setAsset] = useState<Asset | {}>({});
```

web/src/components/organisms/Asset/Asset/hooks.ts
```diff
@@ -0,0 +1,81 @@
...
+  const [selectedPreviewType, setSelectedPreviewType] = useState<PreviewType>(PreviewType.Image);
+  useState<boolean>(false);
```
KaWaite, line 14: This shouldn't be here, right?

web/src/components/organisms/Asset/Asset/hooks.ts
```diff
@@ -0,0 +1,81 @@
...
+          // TODO: notification
+          alert("Asset was successfully updated.");
+        }
+      })(),
+    [updateAssetMutation],
+  );
+
+  useEffect(() => {
+    const asset: Asset = data?.asset as Asset;
+    setAsset(asset);
```
KaWaite, line 46:
```suggestion
    setAsset((data?.asset ?? {}) as Asset );
```
The syntax might be a little off, so please test this

web/src/components/organisms/Asset/Asset/hooks.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (selectedPreviewType === PreviewType.Geo) {
+      viewerRef?.canvas.requestFullscreen();
+    } else {
+      setIsModalVisible(true);
+    }
+  };
+
+  const handleModalCancel = () => {
+    setIsModalVisible(false);
+  };
```
KaWaite, line 69: Please use useCallback for your functions

web/src/utils/sort.ts
```diff
@@ -0,0 +1,9 @@
+import moment from "moment";
+
+type SortFunction<T> = (a: T, b: T) => number;
+
+export const dateSort: SortFunction<Date> = (a, b) => moment(a).diff(moment(b));
+
+export const numberSort: SortFunction<number> = (a, b) => a - b;
+
+export const stringSort: SortFunction<string> = (a, b) => a.localeCompare(b);
```
KaWaite, line 9: I think we talked about this before, but these functions aren't "sorting" are they? Like numberSort is subtracting the first number from the second number, right?

web/src/utils/convert.ts
```diff
@@ -0,0 +1,21 @@
+export const uuidToURL = (uuid: string, filename: string) => {
+  // TODO: needs proper implementation
+  const host = window.REEARTH_CONFIG?.host ?? "http://localhost:8080";
+  const path = getPathFromUUID(uuid, filename);
+  return join(host, path);
+};
+
+export const getPathFromUUID = (uuid: string, filename: string) => {
+  if (!uuid || !filename) {
```
KaWaite, line 9: Your types for uuidToURL and getPathFromUUID don't allow for either param to be undefined or null. If you want to allow that, update the param types. If not, you don't need this if statement

web/src/components/organisms/Asset/AssetList/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+import DownloadButton from "@reearth-cms/components/atoms/DownloadButton";
+import Icon from "@reearth-cms/components/atoms/Icon";
+import {
+  ListToolBarProps,
+  ProColumns,
+  OptionConfig,
+  TableRowSelection,
+  TablePaginationConfig,
+} from "@reearth-cms/components/atoms/ProTable";
+import { UploadProps } from "@reearth-cms/components/atoms/Upload";
```
KaWaite, line 14: Sorry if I have been allowing it up until now in previous PRs, but actually organisms should be made up of ONLY molecules.
Please clean up this and any other organism in this PR that imports atoms

web/src/components/organisms/Asset/AssetList/index.tsx
```diff
@@ -0,0 +1,206 @@
...
+        hideUploadModal();
+      });
+  };
+
+  const uploadProps: UploadProps = {
+    name: "file",
+    multiple: true,
+    directory: false,
+    showUploadList: true,
+    accept: "image/*,.zip,.json,.geojson,.topojson,.shapefile,.kml,.czml,.glb",
```
KaWaite, line 68: Can you move the accepted string outside of the component and export it

web/src/gql/provider.tsx
nourbalaha: Done!
I removed httpLink and I didn't get any type error. maybe apollo-client's newest version fixed the type issue.

web/src/components/organisms/Asset/AssetList/index.tsx
nourbalaha: Done!

web/src/utils/convert.ts
nourbalaha: This if statement also checks if the string is empty.

web/src/components/organisms/Asset/Asset/hooks.ts
nourbalaha: Done!

web/src/components/organisms/Asset/Asset/hooks.ts
nourbalaha: Done!
I tested it and it worked.

web/src/components/organisms/Asset/Asset/hooks.ts
nourbalaha: I got some errors after doing this suggested change.

web/src/components/organisms/Asset/Asset/hooks.ts
nourbalaha: Right, I deleted it.
Done!

web/src/components/organisms/Asset/AssetList/index.tsx
nourbalaha: Done!

web/src/utils/sort.ts
nourbalaha: This is used like this:
```
const numbers = [4,2,6]
const numberSort = (a, b) => a - b;
const sortedNumbers = numbers.sort(numberSort) //[2, 4, 6]
```

web/src/components/organisms/Asset/Asset/hooks.ts
KaWaite: Ok!

web/src/components/organisms/Asset/AssetList/index.tsx
KaWaite: Great!

web/src/utils/sort.ts
KaWaite: Ah now I see where they are being used (passing to an antd component, right?).
Since they are sort method callbacks, not actually doing the sorting, can you update the names to be xxxxSortCallback so the purpose of them is clear.

nourbalaha: Right, I updated the sort callback functions' names to be like this "xxxxSortCallback".
Done!

KaWaite (approved):

PR #153, opened 2022-08-01
feat(server): Model and schema CRUD

yk-eukarya:
what is done:
- Some changes to gql schema
- Some changes to domain models
- Controller layer implemented
- Use case layer implemented
- Repo (mongo) layer implemented


server/pkg/key/key.go
```diff
@@ -32,3 +33,9 @@ func (k Key) IsValid() bool {
 func (k Key) String() string {
 	return k.key
 }
+
+func (k Key) Clone() Key {
+	return Key{
+		key: strings.Clone(k.key),
```
rot1024, line 39:
```suggestion
		key: k.key,
```

string is not a reference, so it does not need to be cloned

rot1024 (changes requested):

server/internal/usecase/interfaces/model.go
```diff
@@ -0,0 +1,38 @@
...
+	Description *string
+	Key         *string
+}
+
+var (
+	ErrModelKey error = errors.New("model key is already used by another model")
+)
+
+type Model interface {
+	Fetch(context.Context, []id.ModelID, *usecase.Operator) (model.List, error)
```
rot1024, line 31:
```suggestion
	FindByIDs(context.Context, []id.ModelID, *usecase.Operator) (model.List, error)
```

I think FindBy~ is better

server/internal/usecase/interfaces/schema.go
```diff
@@ -0,0 +1,40 @@
...
+	Key          *string
+	TypeProperty schema.TypeProperty
+}
+
+var (
+// Err1 error = errors.New("")
+)
+
+type Schema interface {
+	Fetch(context.Context, []id.SchemaID, *usecase.Operator) (schema.List, error)
```
rot1024, line 36:
```suggestion
	FindByIDs(context.Context, []id.SchemaID, *usecase.Operator) (schema.List, error)
```

server/pkg/key/key.go
```diff
@@ -32,3 +32,9 @@ func (k Key) IsValid() bool {
 func (k Key) String() string {
 	return k.key
 }
+
+func (k Key) Clone() Key {
+	return Key{
+		key: k.key,
+	}
+}
```
rot1024, line 40: actually Key.Clone is not needed because it does not include any pointers as a field

rot1024 (approved):

PR #155, opened 2022-08-03
feat(web): update dependencies

issmail-basel:
I changed vite port to always be `3000` to make sure login works (port 3000 is added to Auth0)

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -10,7 +10,7 @@
     "node": ">=16"
   },
   "scripts": {
-    "start": "vite --port 3000",
+    "start": "vite",
     "build": "tsc && vite build",
     "preview": "vite preview",
     "test": "vitest",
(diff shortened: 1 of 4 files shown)


web/package.json
```diff
@@ -10,7 +10,7 @@
     "node": ">=16"
   },
   "scripts": {
-    "start": "vite",
+    "start": "vite --port 3000",
```
rot1024, line 13: Add setting to `vite.config.ts` instead of adding arguments

https://vitejs.dev/config/preview-options.html#preview-port

```ts
export default defineConfig({
  server: {
    port: 3000
  },
  // ...
});
```

rot1024 (changes requested):

KaWaite (approved):

rot1024 (approved):

Issue #156, opened 2022-08-03
[BE]: Replace current implementatin with reearthx pkg


HideBa:
# What
Common processes for building applications are copied and pasted into Re:Earth Backend, Re:Earth CMS Backend, Re:Earth Marketplace, etc. This could increase the cost of maintenance in the future, so we have developed a library called reearthx. In this task, we will replace the current implementation with this reearthx. For example, replacing logger, idx, etc. with reearthx.

# Things to replace
- [x] idx
- [x] log
- [x] ~~authserver~~
- [x] mongo
- [x] rerror
- [x] util
- [x] usecase

To avoid huge PR, plz open the PR when you finish one of these. 🎉

PR #159, opened 2022-08-04
refactor(server): replace log with reearthx

gop50k:
About this ticket
https://github.com/reearth/reearth-cms/issues/156


yk-eukarya (commented):
It's better to remove the unused code in `server/pkg/log/`.

PR #164, opened 2022-08-08
refactor(server): replace idx, log, rerror, util, packages with reearthx

gop50k:
https://github.com/reearth/reearth-cms/issues/156

replaced idx, log, rerror, util pkg. 
and deleted idx, log, rerror, util pkg


yk-eukarya (commented):
LGTM

PR #165, opened 2022-08-08
fix(web): add events and style to icons

issmail-basel:


web/src/components/atoms/Icon/icons.ts
```diff
@@ -44,4 +54,13 @@ export default {
   plus: PlusOutlined,
   close: CloseOutlined,
   exclamationCircle: ExclamationCircleOutlined,
+  textT: Text,
```
KaWaite, line 57: textT?

web/src/components/atoms/Icon/icons.ts
```diff
@@ -44,4 +54,13 @@ export default {
...
+  textAlignLeft: TextAlignLeft,
+  markDown: MarkDown,
```
KaWaite, line 59: markdown

KaWaite (approved):
Just a couple spelling changes

web/src/components/atoms/Icon/icons.ts
issmail-basel: Because it has `T` shape, same name Red used in design file.

PR #174, opened 2022-08-23
feat(web): add schema menu

issmail-basel:
# Overview

## What I've done
Added a new molecule component for showing a list of models, and the ability to create new models.
Also added typescript types for both `model` and `field`.

## How I tested
Tested in the acceptance test

## Screenshot
schema

Syntax issues

### DIFF
--- a/web/src/components/molecules/Schema/SchemaMenu.tsx
+++ b/web/src/components/molecules/Schema/SchemaMenu.tsx
@@ -1,4 +1,5 @@
 import styled from "@emotion/styled";
+import type { MenuInfo } from "rc-menu/lib/interface";
 import React from "react";
 
 import Button from "@reearth-cms/components/atoms/Button";
@@ -23,7 +24,7 @@ const SchemaMenu: React.FC<Props> = ({
   handleModalOpen,
   selectModel,
 }) => {
-  const onClick = (e: any) => {
+  const onClick = (e: MenuInfo) => {
     selectModel(e.key);
   };
 
--- a/web/src/components/molecules/Schema/SchemaMenu.tsx
+++ b/web/src/components/molecules/Schema/SchemaMenu.tsx
@@ -67,7 +67,7 @@ const SchemaStyledMenuTitle = styled.h1`
   font-weight: 400;
   font-size: 14px;
   line-height: 22px;
-  color: rgba(0, 0, 0, 0.45);
+  color: #00000073;
 `;
 
 const SchemaStyledTitle = styled.h2`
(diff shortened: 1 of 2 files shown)


web/src/components/molecules/Schema/types.ts
```diff
@@ -0,0 +1,35 @@
+export type Model = {
+  id: string;
+  name: string;
+  description: string;
```
KaWaite, line 4: Isn't description optional? (In your modalCreationModel PR the form doesn't make description required)

web/src/components/molecules/Schema/types.ts
```diff
@@ -0,0 +1,35 @@
...
+  key: string;
```
KaWaite, line 5: Are these optional? In the modelcreationmodal PR they aren't required, but start with empty string values, so I am a little unsure if that is on purpose and/or desired.

web/src/components/molecules/Schema/SchemaMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+  selectModel: (modelId: string) => void;
+}
+
+const SchemaMenu: React.FC<Props> = ({
+  defaultSelectedKeys,
+  models,
+  handleModalOpen,
+  selectModel,
+}) => {
+  const onClick = (e: any) => {
```
KaWaite, line 26: Can this be typed better to whatever the Menu component's onClick event is?

web/src/components/molecules/Schema/SchemaMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+  };
+
+  return (
+    <SchemaStyledMenu>
+      <SchemaStyledTitle>Schema</SchemaStyledTitle>
+      <SchemaAction>
+        <SchemaStyledMenuTitle>Models</SchemaStyledMenuTitle>
+        <SchemaAddButton
+          onClick={handleModalOpen}
+          color="#1890FF"
```
KaWaite, line 37: You have the color set here and below in the styled components. Pick one and remove the other

web/src/components/molecules/Schema/SchemaMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+    color: #1890ff;
+  }
+`;
+
+const SchemaStyledMenuTitle = styled.h1`
+  margin: 0;
+  font-weight: 400;
+  font-size: 14px;
+  line-height: 22px;
+  color: rgba(0, 0, 0, 0.45);
```
KaWaite, line 74: Can you use hex here since you are using hex everywhere else? Just to keep consistent

KaWaite (commented):
Couple comments

web/src/components/molecules/Schema/types.ts
issmail-basel: @KaWaite , I think key is required, and it is a backend edit that should be done as well

KaWaite (approved):

PR #176, opened 2022-08-23
feat(server): item domain models

mimoham24:
# Overview


server/pkg/item/version.go
```diff
@@ -0,0 +1,56 @@
+package item
+
+import "github.com/reearth/reearth-cms/server/pkg/id"
+
+type Version struct {
+	version string
+	parent  []string
+	ref     []string
+	fields  id.FieldIDList
+}
```
rot1024, line 10: This struct does not hold any information about the value of the field. At least struct like these should be implemented:

```go
type Version struct {
  version version.Version
  parents version.Versions
  refs    version.Refs
  fields  []*Field
}

type Field struct {
  id    FieldID
  typ   schema.Type
  values []*Value // Note that multiple values should be also supproted
}

func NewField(schema.Type, "Value) *Field

type Value struct {
  typ schema.Type
  val any
}

func NewValue(schema.Type, any) *Value
```

Note that separate it to smaller Go files (e.g. `field.go` `value.go`)

rot1024 (changes requested):
`pkg/item_test.go` seems missing

PR #177, opened 2022-08-25
test(server): add test to model memory repo

gop50k:
# Overview


server/internal/infrastructure/memory/model_test.go
```diff
@@ -0,0 +1,146 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/reearth/reearth-cms/server/internal/usecase/repo"
+	"github.com/reearth/reearth-cms/server/pkg/id"
+	"github.com/reearth/reearth-cms/server/pkg/model"
+	"github.com/reearth/reearthx/rerror"
+	"github.com/reearth/reearthx/util"
+	"github.com/reearth/reearthx/usecasex"
```
yk-eukarya, line 12: imports are not sorted. maybe you can use `go-imports-sort` to sort them, I did not try it before since my IDE sorts the imports automatically.

server/internal/infrastructure/memory/model_test.go
```diff
@@ -0,0 +1,146 @@
...
+		endCursor,
+		true,
+		true,
+	)
+	var arg *usecasex.Cursor
+	gotModelList, gotPageInfo, err := r.FindByProject(ctx, pId, usecasex.NewPagination(lo.ToPtr(1), lo.ToPtr(1), arg, arg))
+	assert.NoError(t, err)
+	assert.Equal(t, expectedModelList, gotModelList)
+	assert.Equal(t, expectedPageInfo, gotPageInfo)
+}
```
yk-eukarya, line 64: To cover all the source code you might seed the data with some project. some off code will be executed only if there is returned data.

server/internal/infrastructure/memory/model_test.go
rot1024: You can use gopls (language server) on VSCode to sort imports automatically.

server/internal/infrastructure/memory/model_test.go
```diff
@@ -0,0 +1,146 @@
...
+	ctx := context.Background()
+	pId := id.NewProjectID()
+	r := &Model{
+		data: &util.SyncMap[id.ModelID, *model.Model]{},
+		f:    repo.WorkspaceFilter{},
+	}
+	got, err := r.CountByProject(ctx, pId)
+	assert.NoError(t, err)
+	assert.Equal(t, 0, got)
+}
```
yk-eukarya, line 76: the same here, and also consider adding `models` to different `projects`

yk-eukarya (commented):
We need to increase the coverage now its 57% only, in a file like this one we can make it 100%. the main point is to seed the repo for tests and to test the code within the if branchs.

server/internal/infrastructure/memory/model_test.go
gop50k: @yk-eukarya 
yasser, thanks for your comment.
So you are saying that I should not use `models` unnecessarily in this file??

server/internal/infrastructure/memory/model_test.go
```diff
@@ -0,0 +1,251 @@
...
+	mIds := id.ModelIDList{}
+	r := &Model{
+		data: &util.SyncMap[id.ModelID, *model.Model]{},
+		f:    repo.WorkspaceFilter{},
+	}
+	r.data.Store(mId, model.New().NewID().Schema(id.NewSchemaID()).Key(key).Project(pId).MustBuild())
+
+	expectedModelList := model.List(r.data.FindAll(func(k id.ModelID, m *model.Model) bool {
+		return mIds.Has(k)
+	})).SortByID()
```
rot1024, line 190:
```suggestion
	seed :=  model.New().NewID().Schema(id.NewSchemaID()).Key(key).Project(pId).MustBuild()
	r.data.Store(mId, seed)

	expectedModelList := model.List(seed)
```

Do not aim excessively for DRY in test codes, but give priority to simple, readable code. An expected value is good to be written manually.

Fix others also

rot1024 (changes requested):

PR #180, opened 2022-08-26
feat(server): asset domain models update

nourbalaha:
# Overview
Updated asset domain models

## What I've done
 - used uuid instead of hash.
 - added asset file and asset file builder to domain models.
 - added file to domain models.
 - added sortType to domain models.
 - added more unit tests to increase code coverage.


server/pkg/asset/file_builder.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (b *FileBuilder) Path(path string) *FileBuilder {
+	b.f.path = path
+	return b
+}
+
+func (b *FileBuilder) Children(children []*File) *FileBuilder {
+	b.f.children = slices.Clone(children)
+	return b
+}
```
rot1024, line 53: file_builder.go is not needed. Just initialzie with 

```go
file := &File{
  Content: ...,
  Size: ...,
}
```

server/pkg/asset/asset.go
```diff
@@ -45,13 +47,20 @@ func (a *Asset) Size() uint64 {
 }
 
 func (a *Asset) PreviewType() *PreviewType {
+	if a.previewType == nil {
+		return lo.ToPtr(PreviewType(""))
```
rot1024, line 51:
```suggestion
		return nil
```

server/pkg/asset/asset.go
```diff
@@ -45,13 +47,20 @@ func (a *Asset) Size() uint64 {
...
 }
 
-func (a *Asset) Hash() string {
-	return a.hash
+func (a *Asset) UUID() string {
+	return a.uuid
+}
+
+func (a *Asset) UpdatePreviewType(p *PreviewType) {
+	a.previewType = p
```
rot1024, line 65:
```suggestion
	a.previewType = util.CloneRef(p)
```

Don't just imitate them superficially, but think about why you do it:

```go
pt := lo.ToPtr(PreviewTypeHoge)
asset.UpdatePreviewType(pt)
*pt = PreviewTypeAnother
assert.NotEqual(t, PreviewTypeHoge, *asset.PreviewType())
// The inner field was unexpectedly changed. It means we cannot force developers to force using UpdatePreviewType to update the preview type.
```

rot1024 (changes requested):

server/pkg/asset/asset.go
```diff
@@ -13,7 +15,7 @@ type Asset struct {
 	size        uint64
 	previewType *PreviewType
 	file        *File
```
rot1024, line 17:
```suggestion
```

Asset shoud not have a File. A file has a io.ReadCloser and it cannot be initialized without a connection such as HTTP, and if not released in the correct procedure, it will cause a connection leak. A File struct is only initialized and used in very short time in usecase functions.

rot1024 (changes requested):

server/pkg/asset/asset.go
nourbalaha: File in pkg/asset/file is an AssetFile which has the following attributes:
```type AssetFile {
  name: String!
  size: FileSize!
  contentType: String
  path: String!
  children: [AssetFile!]
}```
it is different from pkg/file.

Comment (rot1024) [reply server/pkg/asset/asset.go:17]: Oh, ok

File: server/pkg/asset/asset_test.go [hunk 5]
```diff
@@ -25,7 +25,7 @@ func TestAsset_AssetType(t *testing.T) {
 		size:        size,
 		previewType: PreviewTypeFromRef(lo.ToPtr(PreviewTypeIMAGE.String())),
 		file:        &File{},
-		hash:        "yyy",
+		uuid:        "xx/xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/hoge",
```
rot1024, line 28:
```suggestion
		uuid:        "xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
```

If you want to get a string as a URL, create it dynamically each time based on the uuid and name; storing it as is in URL format in the DB makes the URL resistant to change.

rot1024 (changes requested):

rot1024 (approved):

PR #185, opened 2022-08-29
feat(server): thread and comment domain models

nourbalaha:
# Overview
In this PR I added domain models for asset's thread and comment


server/pkg/thread/builder.go
```diff
@@ -0,0 +1,42 @@
...
+
+	return b.th, nil
+}
+
+func (b *Builder) MustBuild() *Thread {
+	r, err := b.Build()
+	if err != nil {
+		panic(err)
+	}
+	return r
```
rot1024, line 26:
```suggestion
	return lo.Must(b.Build())
```

server/pkg/thread/comment.go
```diff
@@ -0,0 +1,19 @@
...
+	return c.id
+}
+
+func (c *Comment) Author() UserID {
+	return c.author
+}
+
+func (c *Comment) Content() string {
+	return c.content
+}
```
rot1024, line 19:
```suggestion
}

func (c *Comment) SetContent(content string) string {
	c.content = content
}
```

Add test cases also

server/pkg/asset/asset_test.go
```diff
@@ -37,6 +39,7 @@ func TestAsset_AssetType(t *testing.T) {
 	assert.Equal(t, wantPreviewType, got.PreviewType())
 	assert.Equal(t, &File{}, got.File())
 	assert.Equal(t, "yyy", got.Hash())
+	assert.Equal(t, thid, got.Thread())
 }
```
rot1024, line 43: this test function has not good name: `TestAsset_AssetType` because it no longer tests only AssetType method, but it tests all getter methods. so rename it, e.g. `TestAsset` `TestAsset_Getters`

rot1024 (changes requested):

server/pkg/thread/builder_test.go
```diff
@@ -55,49 +56,19 @@ func TestBuilder_Build(t *testing.T) {
...
-					MustBuild()
-			}
-			if tt.err != nil {
-				assert.PanicsWithValue(t, tt.err, func() { _ = build() })
-			} else {
-				assert.Equal(t, tt.want, build())
-			}
-		})
-	}
+	assert.Equal(t, want, got)
```
rot1024, line 71: Add an abnormal test case that Must method panics.

To get an idea of how many test cases you need, consider increasing code coverage. That means eliminating as many untested lines as possible!

server/pkg/thread/comment_test.go
```diff
@@ -10,14 +11,30 @@ func TestComment_CommentType(t *testing.T) {
...
 	assert.Equal(t, c, got.Content())
+	assert.Equal(t, tim, got.CreatedAt())
+}
+
+func TestComment_SetContent(t *testing.T) {
+	c := "xxx"
+	got := Comment{}
+
+	got.SetContent(c)
+	assert.Equal(t, got.Content(), c)
```
rot1024, line 34:
```suggestion
	comment := Comment{}
	comment.SetContent(c)
	assert.Equal(t, "xxx", comment.content)
```

- "got" = "result". In this case it is not proper name.
- Note the order: `assert.Equal(t, EXPECTED, ACTUAL)`
- When testing only SetContent, it is best to use as few other methods as possible. When the implementation of other methods is changed, this test case will also be affected.

server/pkg/thread/comment_test.go
```diff
@@ -10,14 +11,30 @@ func TestComment_CommentType(t *testing.T) {
...
+}
+
+func TestComment_CreatedAt(t *testing.T) {
+	var got *Comment = nil
+	assert.Equal(t, time.Time{}, got.CreatedAt())
```
rot1024, line 39:
```suggestion
	var got *Comment = nil
	assert.Equal(t, time.Time{}, got.CreatedAt())
	got = &Comment{ID: NewID()}
	assert.Equal(t, got.id.Timestamp(), got.CreatedAt())
```

Please add more test cases as they are missing.

server/pkg/thread/thread_test.go
```diff
@@ -8,14 +8,17 @@ import (
 
 func TestThread_ThreadType(t *testing.T) {
```
rot1024, line 9:
```suggestion
func TestThread_Getters(t *testing.T) {
```

As I pointed out before.

`TestSTRUCTNAME_METHODNAME(t *testing.T)`

rot1024 (changes requested):

server/pkg/thread/thread_test.go
nourbalaha: Done!

server/pkg/thread/comment_test.go
nourbalaha: Done!

server/pkg/thread/builder_test.go
nourbalaha: Done!

server/pkg/thread/comment_test.go
nourbalaha: Done!

rot1024 (approved):

server/pkg/thread/thread_test.go
```diff
@@ -0,0 +1,34 @@
...
+	assert.Equal(t, thid, got.ID())
+	assert.Equal(t, wid, got.Workspace())
+	assert.Equal(t, c, got.Comments())
+}
+
+func TestThread_Comments(t *testing.T) {
+	var got *Thread = nil
+	assert.Nil(t, got.Comments())
+
+	c := []*Comment{}
```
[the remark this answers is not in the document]
nourbalaha: Done!

server/pkg/thread/comment.go
```diff
@@ -0,0 +1,34 @@
...
+
+func (c *Comment) CreatedAt() time.Time {
+	if c == nil {
+		return time.Time{}
+	}
+
```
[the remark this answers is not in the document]
nourbalaha: Done!

server/pkg/thread/comment.go
```diff
@@ -0,0 +1,34 @@
+package thread
+
+import "time"
+
+type Comment struct {
+	id        CommentID
+	author    UserID
+	content   string
+	createdAt time.Time
```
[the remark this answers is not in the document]
nourbalaha: I know that id already have a timestamp, however I think that we still need createdAt in the Comment struct right?

PR #186, opened 2022-08-29
feat(server): thread and comment schema

nourbalaha:
# Overview
In this PR I added a schema for thread and comment.


server/schemas/thread.graphql
```diff
@@ -0,0 +1,35 @@
+type Thread {
+  id: ID!
+  comments: [Comment]
```
rot1024, line 3:
```suggestion
  workspaceId: ID!
  comments: [Comments!]!
  workspace: Workspace
```

workspace should be a "resolver"

server/schemas/thread.graphql
```diff
@@ -0,0 +1,35 @@
...
+  content: String!
+}
+
+input CreateCommentInput {
+  threadId: ID!
+  authorId: ID!
+  content: string!
+}
+
+input UpdateCommentInput {
```
rot1024, line 18:
```suggestion
input UpdateCommentInput {
  threadId: ID!
```

server/schemas/thread.graphql
```diff
@@ -0,0 +1,35 @@
...
+  commentId: ID!
+  content: string!
+}
+
+input DeleteCommentInput {
+  CommentId: ID!
```
rot1024, line 24:
```suggestion
  threadId: ID!
  commentId: ID!
```

server/schemas/thread.graphql
```diff
@@ -0,0 +1,35 @@
...
+}
+
+extend type Query {
+  comments(threadId: ID!): [Comment]
+}
```
rot1024, line 29:
```suggestion
```

I think it will not be used

server/schemas/thread.graphql
```diff
@@ -0,0 +1,35 @@
...
+
+extend type Mutation {
+  createComment(input: CreateCommentInput!): Comment: Comment!
+  UpdateComment(input: UpdateCommentInput!): Comment: Comment!
```
rot1024, line 33:
```suggestion
  updateComment(input: UpdateCommentInput!): Comment: Comment!
```

server/schemas/thread.graphql
```diff
@@ -0,0 +1,35 @@
+type Thread {
+  id: ID!
+  comments: [Comment]
+}
+
+type Comment {
+  id: ID!
+  authorId: ID!
+  content: String!
```
rot1024, line 9:
```suggestion
  content: String!
  createdAt: DateTime!
```

createdAt can use a timestamp of a commnet ID, so there is no need to add a `createdAt` field to the domain model.

rot1024 (changes requested):
thread.graphql -> comment.graphql

server/schemas/thread.graphql
```diff
@@ -0,0 +1,49 @@
...
+  id: ID!
+  author: User
+  authorId: ID!
+  content: String!
+  createdAt: DateTime!
+}
+
+input CreateCommentInput {
+  threadId: ID!
+  authorId: ID!
```
yk-eukarya, line 18: Do we need the `authorId` here? cant we use the logged-in user id from `operator`

nourbalaha: @yk-eukarya I think that we do not need authorId here, this applies to `UpdateCommentInput` as well right?

nourbalaha: Done!

rot1024 (approved):

PR #190, opened 2022-08-29
feat(web): asset's atoms updates

nourbalaha:
In this PR I added the following:
1- Changed the way downloadButton works.
2- Added export types that are related to proTable.
3- Added new icons.


web/src/components/atoms/DownloadButton/index.tsx
```diff
@@ -19,7 +18,20 @@ const DownloadButton: React.FC<DownloadButtonProps> = ({
   ...props
 }) => {
   const handleDownload: MouseEventHandler<HTMLElement> | undefined = () => {
-    saveAs(url, filename);
+    fetch(url, {
```
KaWaite, line 21: Can you explain why you stopped using file-saver and are kinda doing it manually now?

web/src/components/atoms/ProTable/index.tsx
```diff
@@ -1,5 +1,13 @@
...
+import { TableRowSelection } from "@ant-design/pro-table/lib/typing";
+import { TablePaginationConfig } from "antd";
 
 export default ProTable;
 export type { ProTableProps };
+export type { ListToolBarProps };
+export type { ProColumns };
+export type { OptionConfig };
+export type { TableRowSelection };
+export type { TablePaginationConfig };
```
KaWaite, line 13: You can export all types at once
```
export type { 
     ProTableProps, 
     ListToolBarProps, 
     etc, 
     etc};
```

web/src/components/atoms/ProTable/index.tsx
```diff
@@ -1,5 +1,13 @@
-import ProTable from "@ant-design/pro-table";
-import type { ProTableProps } from "@ant-design/pro-table";
+import ProTable, { ListToolBarProps } from "@ant-design/pro-table";
+import type { ProTableProps, ProColumns } from "@ant-design/pro-table";
```
KaWaite, line 2: Is there a reason ListToolBarProps isn't imported with ProTableProps and ProColumns

KaWaite (commented):
Please see the comments

web/src/components/atoms/DownloadButton/index.tsx
nourbalaha: When I try to download an image with file-saver, it gets opened in a new tab rather than downloading.

KaWaite: In reearth-web I use this [js-file-download package](https://github.com/eligrey/FileSaver.js) which works well if all you need is to download the specified file to the user's computer!

nourbalaha: [js-file-download package](https://github.com/eligrey/FileSaver.js) is the same as file-saver.
it worked on my computer in the past, but recently for some reason it stopped working and started to open files in new tabs instead of downloading them.
if you think that file-saver is reliable, I will revert the changes that I have done to the DownloadButton component and keep using file-saver instead.

nourbalaha: I used js-file-downoad instead of file-saver

KaWaite (approved):

PR #195, opened 2022-08-30
feat(web): add schema hooks

issmail-basel:
# Overview
GraphQL hooks for schema


web/src/components/organisms/Project/Schema/hooks.ts
```diff
@@ -0,0 +1,172 @@
...
+};
+
+export default ({ projectId, modelId }: Params) => {
+  const [modelModalShown, setModelModalShown] = useState(false);
+  const [fieldModalShown, setFieldModalShown] = useState(false);
+
+  const { data, refetch } = useGetModelsQuery({
+    variables: { projectId: projectId ?? "", first: 100 },
+    skip: !projectId,
+  });
```
KaWaite, line 24: Are we paginating? This looks like how reearth-web used to look when we didn't paginate and when pagination was a little broken.

web/src/components/organisms/Project/Schema/hooks.ts
```diff
@@ -0,0 +1,172 @@
...
+                })),
+              },
+            }
+          : undefined,
+      )
+      .filter((model): model is Model => !!model);
+  }, [data?.models.nodes]);
+
+  const rawModel = useMemo(
+    () => data?.models.nodes.find((p: any) => p?.id === modelId),
```
KaWaite, line 53: Can't p be typed better than any?

web/src/components/organisms/Project/Schema/hooks.ts
```diff
@@ -0,0 +1,172 @@
...
+        : undefined,
+    [rawModel],
+  );
+
+  const [createNewModel] = useCreateModelMutation({
+    refetchQueries: ["GetModels"],
+  });
+
+  const handleModelCreate = useCallback(
+    async (data: { name: string; description: string; key: string }) => {
```
KaWaite, line 85: For both this function and handleFieldCreate is there a good reason you are making the param an object? Since the mutations don't require an object isn't better to make the callback's params the same, then you can pass them in more simply

issmail-basel: @KaWaite , passing like this is much easier because of ant.d forms helpers, so I don't need to change on the component part.

web/src/components/organisms/Project/Schema/hooks.ts
issmail-basel: @KaWaite , it is related to generated files, could not find suitable typing.

web/src/components/organisms/Project/Schema/hooks.ts
```diff
@@ -0,0 +1,172 @@
...
+          typeProperty: data.typeProperty,
+        },
+      });
+      if (field.errors || !field.data?.createField) {
+        setModelModalShown(false);
+        return;
+      }
+
+      setModelModalShown(false);
+      refetch();
```
KaWaite, line 141: You refetch here, but you already are doing refetchQueries on your mutations, so I think you are doing it twice (or once after an error even though you seem to not want to)

web/src/components/organisms/Project/Schema/hooks.ts
KaWaite: But you seem to know what p is (here you check if its id is same as modelId and then you use the rawModel to create the modal below) so  can't you create a model type  that meets your expectations and the data you are receiving?

KaWaite: Also, `p` doesn't seem like it fits. If it was a plugin, that makes sense. If they are models `m` or `model` is probably clearer

KaWaite (approved):

PR #200, opened 2022-08-31
feat(web): add field list

issmail-basel:
# Overview
Add field list component

## What I've done
Define fields type, show field list UI


web/src/components/molecules/Schema/FieldList.tsx
```diff
@@ -0,0 +1,101 @@
+import styled from "@emotion/styled";
+import React from "react";
+
+import Icon from "@reearth-cms/components/atoms/Icon";
+import List from "@reearth-cms/components/atoms/List";
+
+import { fieldTypes } from "./fieldTypes";
+import { FieldType } from "./types";
```
KaWaite, line 8: Can you put these in the same file?

KaWaite (approved):
Approved BUT if you can combine the types into the same file that would be nicer since they are related!

web/src/components/molecules/Schema/FieldList.tsx
```diff
@@ -0,0 +1,101 @@
...
+  {
+    title: "URL",
+    fields: ["URL"],
+  },
+];
+
+const FieldList: React.FC<Props> = ({ addField }) => {
+  return (
+    <>
+      <h1>Add Field</h1>
```
KaWaite, line 43: Don't need to use the Typography antd component?

PR #204, opened 2022-08-31
fix(server): gql generated code and schema filter 

yk-eukarya:
## Overview
While preparing for the acceptance test multi-issue appeared, this PR is for fixing them. the list below of all changes.

## What I've done
- fix graphQL generated code to reflect the schema
- add workspaceFilter for the schema usecase
- fix the update field mutation (type property was not implemented!)


server/internal/infrastructure/mongo/schema.go
```diff
@@ -33,9 +33,13 @@ func (r *schemaRepo) init() {
...
 
 func (r *schemaRepo) Filtered(f repo.WorkspaceFilter) repo.Schema {
+	filter := f
+	if r.f != nil {
+		filter = f.Merge(filter)
+	}
 	return &schemaRepo{
 		client: r.client,
-		f:      lo.ToPtr(r.f.Merge(f)),
+		f:      &filter,
```
rot1024, line 42: I think the previous way of writing was more immediate.

If an error occurs, there is a problem with the implementation of the Merge method.

rot1024 (changes requested):

server/internal/infrastructure/mongo/schema.go
yk-eukarya: What if `r.f` is not set and equal to `nil`? the `Merge` method is a method on `WorkspaceFilter` not on `*WorkspaceFilter` and I can not check if it is a `nil`!

rot1024: OK so you can fix it as following:

```go
type Hoge struct {
    client *mongox.ClientCollection
	f    repo.WorkspaceFilter // remove "*"
}

func (r *Hoge) Filtered(f repo.WorkspaceFilter) repo.Hoge {
	return &Hoge{
		data: r.data,
		f:    r.f.Merge(f),
	}
}
```

yk-eukarya: In this case, the `workspaceFilter` is required. is that ok? I was thinking of it as optional.

rot1024: Refer to https://github.com/reearth/reearth/blob/main/server/internal/infrastructure/mongo/asset.go#L22

If the contents of WorkspaceFilter are empty, it is treated as if the filter is not applied, so there is no problem.

rot1024 (approved):

PR #205, opened 2022-09-01
feat(worker): unzip compressed file with buffer

HideBa:
# Overview
Implemented the basic logic to package to unzip the zip file. I'll implement/fix other layers such as `controller`, `handler`, etc. 

## What I've done
* `pkg/zip`: a core package that unzips the zip file and writes unzipped data to somewhere. The Zip file package requires an internal io.ReaderAt. However, GCS does not provide those interfaces, so we wrapped the GCS object to be able to provide io.ReaderAt. ReaderAt. Also, to prevent performance degradation due to the large amount of IO being executed, we wrapped it so that it is buffered in memory to some extent. In addition, unzipped files are spread across multiple locations, and their names are not known until they are decompressed. Therefore, this package has an internal "pass a filename to receive a function to upload that file" function, which can dynamically generate upload functions.

* `usecase`: get the actual zip file from gateway layer, calls `pkg/zip` 
* added other layers for application such as `controller`, `app`, `handler`, etc
* CI for worker

## What I haven't done
* the logic to notify CMS server that zip file has been unzipped and uploaded to storage
* 
## How I tested
* I wrote a Unit test that creates temporary text files locally and zip files from them and verifies that they can be extracted properly. This ensures that we only test the important parts without relying on external sources such as GCS.

* One particular point I would like you to review is the `pkg/zip` package and the `infrastructure/gcs` package. The other layers do not need to be looked at in that much detail as they are still partially in the process of being implemented. We plan to revise those again later. I'll highlight files that I want you to review.


worker/internal/infrastructure/gcp/gcs.go
```diff
@@ -0,0 +1,161 @@
...
+	cacheSize               = 10 * 1024 * 1024
+)
+
+type fileRepo struct {
+	bucketName   string
+	base         *url.URL
+	cacheControl string
+}
+
+func NewFile(bucketName, base string, cacheControl string) (gateway.File, error) {
```
HideBa, line 30: Plz review this file well

worker/internal/usecase/interactor/decompress.go
```diff
@@ -0,0 +1,50 @@
...
+
+type Usecase struct {
+	gateways *gateway.Container
+}
+
+func NewUsecase(g *gateway.Container) *Usecase {
+	return &Usecase{gateways: g}
+}
+
+func (u *Usecase) Decompress(ctx context.Context, assetURL string) error {
```
HideBa, line 20: Plz review this function.

worker/pkg/zip/zip.go
```diff
@@ -0,0 +1,48 @@
...
+	"archive/zip"
+	"io"
+)
+
+type Unzipper struct {
+	r   *zip.Reader
+	wFn func(name string) (io.Writer, error)
+}
+
+func NewUnzipper(ra io.ReaderAt, size int64, wFn func(name string) (io.Writer, error)) (*Unzipper, error) {
```
HideBa, line 13: Plz review this file well

worker/internal/app/app.go
```diff
@@ -1,15 +1,87 @@
...
+	e := echo.New()
+	e.Debug = cfg.Debug
+	e.HideBanner = true
+	e.HidePort = true
+	e.HTTPErrorHandler = errorHandler(e.DefaultHTTPErrorHandler)
+
+	// basic middleware
+	logger := rlog.NewEcho()
+	e.Logger = logger
+	e.Use(middleware.Recover(), otelecho.Middleware("reearth-cms"))
```
rot1024, line 30:
```suggestion
	e.Use(middleware.Recover(), otelecho.Middleware("reearth-cms/worker"))
```

worker/internal/app/config.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+const configPrefix = "reearth"
+
+type Config struct {
+	Port         string `default:"8080" envconfig:"PORT"`
+	ServerHost   string
+	Dev          bool
+	Host_Web     string
+	AssetBaseURL string `default:"http://localhost:8080/assets"`
```
rot1024, line 19:
```suggestion
```

worker/internal/app/main.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/labstack/echo/v4"
+
+	rhttp "github.com/reearth/reearth-cms/worker/internal/adapter/http"
+	"github.com/reearth/reearth-cms/worker/internal/usecase/gateway"
+	"github.com/reearth/reearth-cms/worker/internal/usecase/interactor"
+	"github.com/reearth/reearthx/log"
+)
+
+func Start(debug bool, version string) {
+	log.Infof("reearth-cms %s", version)
```
rot1024, line 17:
```suggestion
	log.Infof("reearth-cms/worker %s", version)
```

worker/internal/app/app.go
```diff
@@ -1,15 +1,87 @@
...
+
+	api := e.Group("/api")
+
+	api.GET("/ping", func(c echo.Context) error { return c.JSON(http.StatusOK, "pong") })
+
+	t := handler.DecompressHandler()
+	api.GET("/decompress", t)
```
rot1024, line 37:
```suggestion
```

GET should not cause side effects.

worker/internal/usecase/gateway/file.go
```diff
@@ -0,0 +1,22 @@
...
+	ErrFailedToUploadFile error = errors.New("failed to upload file")
+	ErrFileTooLarge       error = errors.New("file too large")
+	ErrFailedToRemoveFile error = errors.New("failed to remove file")
+)
+
+type File interface {
+	// ReadAsset(ctx context.Context, name string) (io.ReadCloser, error)
+	RandomReadAssetByURL(ctx context.Context, url *url.URL) (io.ReaderAt, int64, error)
+	UploadAssetFunc(ctx context.Context, name string) (io.WriteCloser, error)
+}
```
rot1024, line 22:
```suggestion

type ReadAtCloser interface {
	io.ReaderAt
	io.Closer
}

type File interface {
    Read(ctx context.Context, path string) (io.ReadAtCloser, int64, error)
	Upload(ctx context.Context, name string) (io.WriteCloser, error)
}
```

- There is no way to disconnect, causing connection leaks. io.ReaderAt with io.Closer would be needed.
- File does not have any knowledge of Asset. Therefore, it is unnatural for Asset to be used as a method name in File.
- Simpler method names. Imagene usage: `file.Upload()`. This is enough to know what is being done.
- `Read` should accept a path string instead of a URL. File has no knowledge of where it exists in the first place. It does not know whether it is in the GCS or a local file. It is only interested in the its own file system that exists somewhere. And the only way to express the location of a file in the file system is by path. If the File accept URLs, it have to deal with extra information about where acually it is.

worker/pkg/zip/zip_test.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	testFilePrefix = "test"
+	fileContent    = "Hello"
+)
+
+// createDummyFiles creates dummy files with given filenames
+func createDummyFiles(t *testing.T, fileNames ...string) {
```
rot1024, line 22: Unzipper only needs io.ReadAt, so bytes.Buffer is sufficient without creating a file. Creating a file adds an extra external factor that prevents a successful test, so if you don't need it, refrain from doing so.

Also, although not applicable in this case, afero.FS can be substituted in many cases if you wish to test emulating a file system.

worker/pkg/zip/zip.go
```diff
@@ -0,0 +1,48 @@
...
+			if err != nil {
+				return err
+			}
+			defer rc.Close()
+
+			w, err := uz.wFn(f.FileInfo().Name())
+			if err != nil {
+				return err
+			}
+			_, err = io.Copy(w, rc)
```
rot1024, line 40: Zip bomb protection is needed.

```suggestion
			_, err = io.CopyN(w, rc, limit)
			_ = w.Close()
			if errors.Is(io.EOF, err) {
				return &LimitError{ Path: f.FileInfo().Name() }
			}
```

with 

```go
const limit = 1024 * 1024 * 1024 * 10 // 10GB

type LimitError struct {
  Path string
}

func (e *LimitError) Error() string {
  return fmt.Sprtinf("file size limit reached at %s", e.Path)
}
```

worker/pkg/zip/zip.go
rot1024, line 13:
```suggestion
func NewUnzipper(r *zip.Reader, wFn func(name string) (io.WriteCloser, error)) (*Unzipper, error) {
```

- If a new parameter is added to `zip.NewReader`, `NewUnzipper` will also be affected. In these cases, there is a mixture of multiple interests.
- If it does not have the ability to close a Writer, users will have to save all the Writers they create in some other variable to ensure the Writer is closed.

rot1024 (changes requested):
Add `.env` and `.env.*` to `.gitignore` at the root

worker/internal/app/app.go
HideBa: Opps, sorry I forgot to change it to POST since it's easier to Debug with using GET. I'll fix it!

worker/internal/usecase/gateway/file.go
HideBa: >Read should accept a path string instead of a URL. File has no knowledge of where it exists in the first place. It does not know whether it is in the GCS or a local file. It is only interested in the its own file system that exists somewhere. And the only way to express the location of a file in the file system is by path. If the File accept URLs, it have to deal with extra information about where acually it is.

hmm, I see, I agree with that since the hostname of them can also be changed, and using the path is more flexible for the change.  

Question: Why Re:Earth uses `url.URL` [these part](https://github.com/reearth/reearth/blob/c7e42351b0af901dd6060f3555c869e7103929a0/server/internal/usecase/gateway/file.go#L22)? Are these expected to refactored in the future?

worker/pkg/zip/zip.go
HideBa: OK! 👍

worker/pkg/zip/zip_test.go
HideBa: I've fixed [refine test case](https://github.com/reearth/reearth-cms/pull/205/commits/03ceb2728dd414243d62a87796fbfd5024f7c198)

worker/pkg/zip/zip_test.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// Buffer implements io.ReadAtCloser and io.WriteCloser
+type Buffer struct {
+	bytes.Buffer
+}
+
+func (*Buffer) Close() error {
+	return nil
+}
```
HideBa, line 22: @rot1024 
Thanks for reviewing my PR. Basically, I accepted your suggested change. 
Additionally, I changed `pkg/zip` for its testing way so that is doesn't need to generate files and zip files dynamically.

worker/pkg/zip/zip.go
```diff
@@ -0,0 +1,60 @@
...
+			if err != nil {
+				return err
+			}
+			_, err = io.CopyN(w, rc, limit)
+			_ = w.Close()
+			if errors.Is(io.EOF, err) {
+				continue
+			}
+			if err != nil {
+				return &LimitError{Path: f.FileInfo().Name()}
```
HideBa, line 46: I also  a little changed your suggestion since if `err` is `io.EOF` it shouldn't return error, just `continue` the process.

HideBa (commented):
Kindly review where I've changed.

worker/pkg/zip/zip_test.go
```diff
@@ -0,0 +1,68 @@
...
+	require.NoError(t, err)
+	uz, err := NewUnzipper(r, wFn)
+	require.NoError(t, err)
+
+	log.Default().Print(expectedFiles, uz)
+	err = uz.Unzip()
+	assert.NoError(t, err)
+	for k, v := range files {
+		assert.Equal(t, v.Bytes(), expectedFiles[k].Bytes())
+	}
```
rot1024, line 66:
```suggestion
	expectedFiles := map[string][]byte{
		"test1.txt": []byte("hello1"),
		"test2.txt": []byte("hello2"),
	}

	// map of buffers which will keep unzipped data
	files := map[string]*Buffer{
		"test1.txt": {bytes.Buffer{}},
		"test2.txt": {bytes.Buffer{}},
	}

	r, err := zip.NewReader(zf, fInfo.Size())
	require.NoError(t, err)
	uz, err := NewUnzipper(r,  func(name string) (io.WriteCloser, error) {
		return files[name], nil
	})
	require.NoError(t, err)

	assert.NoError(t, uz.Unzip())
	for k, v := range files {
		assert.Equal(t, v, expectedFiles[k].Bytes())
	}
```

worker/internal/app/app.go
```diff
@@ -1,15 +1,86 @@
...
+	e := echo.New()
+	e.Debug = cfg.Debug
+	e.HideBanner = true
+	e.HidePort = true
+	e.HTTPErrorHandler = errorHandler(e.DefaultHTTPErrorHandler)
+
+	// basic middleware
+	logger := rlog.NewEcho()
+	e.Logger = logger
+	e.Use(middleware.Recover(), otelecho.Middleware("reearth-cms/worker"))
```
rot1024, line 30:
```suggestion
	e.Use(
		middleware.Recover(),
		logger.AccessLogger(),
		otelecho.Middleware("reearth-cms/worker"),
	)
```

worker/internal/usecase/interactor/decompress.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return err
+	}
+	unzipper, err := rzip.NewUnzipper(zr, uploadFunc)
+	if err != nil {
+		return err
+	}
+
+	err = unzipper.Unzip()
+	if err != nil {
```
rot1024, line 50:
```suggestion
	if err := unzipper.Unzip(); err != nil {
```

worker/pkg/zip/zip_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+}
```
rot1024, line 68: It is also better to test if the Unzip method itself returns the same error when writerFn returns an error. 

Also, to test if the zip bomb attack countermeasure is effective, we could combine a Reader that spits out the same value forever with a Writer that discards the value (optional, as it increases the difficulty a bit).

rot1024 (changes requested):

worker/pkg/zip/zip_test.go
HideBa: fixed wtih [add test to check error](https://github.com/reearth/reearth-cms/pull/205/commits/e2a2d668e41f76afc910c22cff6286eb0892eba3)

HideBa: Since I reversed expected value and actual value, I've fixed it with
[reverse expected and actual](https://github.com/reearth/reearth-cms/pull/205/commits/71ec0544f2445f9809928650b9000e3f33f50b1f)

worker/pkg/zip/zip_test.go
```diff
@@ -0,0 +1,79 @@
...
+		"test2.txt": []byte("hello2"),
+	}
+
+	// map of buffers which will keep unzipped data
+	files := map[string]*Buffer{
+		"test1.txt": {bytes.Buffer{}},
+		"test2.txt": {bytes.Buffer{}},
+	}
+
+	//Normal Scenario
```
rot1024, line 59:
```suggestion
	// Normal Scenario
```

worker/pkg/zip/zip_test.go
```diff
@@ -0,0 +1,79 @@
...
+		return files[name], nil
+	})
+	require.NoError(t, err)
+
+	assert.NoError(t, uz.Unzip())
+	for k, v := range files {
+		assert.Equal(t, v.Bytes(), expectedFiles[k])
+	}
+
+	//Exception: test if  wFn's error is same as what Unzip returnsn
```
rot1024, line 72:
```suggestion
	// Exception: test if wFn's error is same as what Unzip returns
```

worker/pkg/zip/zip_test.go
```diff
@@ -0,0 +1,79 @@
...
+	uz, err = NewUnzipper(r, func(name string) (io.WriteCloser, error) {
+		return nil, errors.New("test")
+	})
+	require.NoError(t, err)
+	assert.Equal(t, errors.New("test"), uz.Unzip())
+
```
rot1024, line 78:
```suggestion
```

worker/pkg/zip/zip_test.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	r, err := zip.NewReader(zf, fInfo.Size())
+	assert.NoError(t, err)
+	_, err = NewUnzipper(r, wFn)
+	assert.NoError(t, err)
+}
+
+func TestUnzipper_Unzip(t *testing.T) {
+	zf, err := os.Open("testdata/test.zip")
+	require.NoError(t, err)
```
rot1024, line 41:
```suggestion
	assert.NoError(t, err)
```

rot1024 (approved):

PR #207, opened 2022-09-02
fix(web): pass select field values

issmail-basel:
# Overview
I'm passing select field values instead of the whole form to keep data up-to-date


web/src/components/molecules/Schema/FieldModal/FieldUpdateModal/index.tsx
```diff
@@ -47,6 +48,7 @@ const FieldUpdateModal: React.FC<Props> = ({
 }) => {
   const [form] = Form.useForm();
   const { TabPane } = Tabs;
+  const selectValues = Form.useWatch("values", form);
```
KaWaite, line 51: Does `currentValues` make more sense here? If `select` makes more sense to you I think this needs to be `selectedValues`

issmail-basel: @KaWaite , yup you are right.
`selectedValues` will be, at it is related to select field only in this file

KaWaite (commented):
Just a single comment about naming

KaWaite (approved):

PR #209, opened 2022-09-04
fix(server): updating the schema type property

yk-eukarya:
# Overview


server/pkg/schema/type_property.go
```diff
@@ -230,6 +230,37 @@ func (t *TypeProperty) Match(m TypePropertyMatch) {
 	}
 }
 
+func (t *TypeProperty) UpdateFrom(t2 *TypeProperty) {
+	if t == nil {
+		return
+	}
+	if t.text != nil && t2.text != nil {
```
rot1024, line 237: param.TypeProperty.Match is not used?

rot1024 (commented):
I think it's better to implement TypeProperty.Clone() and Field.SetTypeProperty() .

```go
f.SetTypeProperty(&param.TypeProperty)

function (f *Field) SetTypeProperty(typeProperty *TypeProperty) {
  f.typeProperty = typeProperty.Clone()
}
```

This allows for a clearer division of responsibilities. This means that the Field does not need to have any knowledge of the
TypeProperty (except for calling the Clone method). This way, when the Match method of a TypeProperty is changed, there is no need to change the Field at all. （

server/pkg/schema/type_property.go
```diff
@@ -230,6 +230,59 @@ func (t *TypeProperty) Match(m TypePropertyMatch) {
 	}
 }
 
+func (t *TypeProperty) UpdateFrom(t2 *TypeProperty) {
```
rot1024, line 233: Since the processing is almost the same as for `Clone` method, it is a duplicate code and should be removed.

rot1024 (approved):

PR #225, opened 2022-09-09
feat(web): improve loading page

issmail-basel:
# Overview
When logging in, the loading identicator is not clear.

## What I've done
Added a spinner in the center of the page when loading backend requests after logging in.


KaWaite (commented):
This looks good, but was it chosen/approved by @lavalse ?

web/src/components/pages/RootPage/index.tsx
```diff
@@ -18,7 +20,11 @@ const RootPage: React.FC = () => {
     }
   }, [isAuthenticated, navigate, login, isLoading]);
 
-  return isLoading ? <h1>{t("Loading")}</h1> : null;
+  return isLoading ? (
+    <Row justify="center" align="middle" style={{ minHeight: "100vh" }}>
+      <Spin tip={t("Loading")} size="large" />
+    </Row>
```
KaWaite, line 26: Can you at least make this in to a molecule and make it a little more reusable? 
I feel like it'd be useful having a Loading molecule similar to this. You'd need to change the 100vh probably to make it take up 100% of its parent or something so it could be used in many locations

issmail-basel: @KaWaite , no need for the `loading`, it is a ready antd component I just added the row to make it full screen and centered.
I put the whole code in a molecules to use it if we need a full page loading, otherwise `loading` antd atom is enough

KaWaite (approved):

PR #241, opened 2022-09-20
feat(web): asset ui fixes

nourbalaha:
# Overview
In this PR I fixed some UI related issues in asset and assetList pages

## What I've done
- Fixed uuidToURL function to work with both file storage and GCS.
- Fixed download asset.
- Username instead of id in CreatedByID.
- Added convertAsset function to convert from GQLAsset to local Asset type.
- Changed PreviewType type from enum to a normal type.
- Moved AssetSortType into a separate file.
- Updated the asset schema to replace `createdById: string` with `createdBy: User`

## What I haven't done
- Pagination.
- Sort


web/src/components/atoms/DownloadButton/index.tsx
```diff
@@ -19,7 +19,16 @@ const DownloadButton: React.FC<DownloadButtonProps> = ({
   ...props
 }) => {
   const handleDownload: MouseEventHandler<HTMLElement> | undefined = () => {
-    fileDownload(url, filename);
+    fetch(url, {
```
issmail-basel, line 22: @nourbalaha , maybe use `async/await` while fetching API to keep the codebase consistent?

web/src/components/atoms/DownloadButton/index.tsx
```diff
@@ -19,7 +19,16 @@ const DownloadButton: React.FC<DownloadButtonProps> = ({
...
+      method: "get",
```
issmail-basel, line 23: method `get` should be upper-case by convention. (to be `GET`)

web/src/components/molecules/Asset/Asset/AssetBody/svgPreview.tsx
```diff
@@ -7,7 +7,12 @@ const SVGPreview: React.FC<Props> = ({ url, svgRender }) => {
   const [svgText, setSvgText] = useState("");
 
   const fetchData = useCallback(() => {
-    fetch(url)
+    fetch(url, {
+      method: "get",
```
issmail-basel, line 11: method `get` should be upper-case by convention. (to be `GET`)

web/src/components/molecules/Asset/AssetList/AssetListTable/index.tsx
```diff
@@ -75,13 +74,13 @@ const AssetListTable: React.FC<AssetListTableProps> = ({
       title: "Created At",
       dataIndex: "createdAt",
       key: "createdAt",
-      sorter: (a, b) => dateSortCallback(a.createdAt, b.createdAt),
+      sorter: (a, b) => dateSortCallback(new Date(a.createdAt), new Date(b.createdAt)),
```
issmail-basel, line 77: I'm just curious, what scenario did you face to have this change?

web/src/components/organisms/Asset/AssetList/assetSortType.type.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type AssetSortType = "date" | "name" | "size";
+
+const enumTypeMapper: Partial<Record<GQLSortType, string>> = {
+  [GQLSortType.Date]: "date",
+  [GQLSortType.Name]: "name",
+  [GQLSortType.Size]: "size",
+};
+
+export function toGQLEnum(val?: AssetSortType) {
```
issmail-basel, line 11: Is this function necessary?
Maybe type assertion (using as) is enough when needed like this:
`sort: sort?.type as GQLSortType,`

issmail-basel (commented):
Please check the comments.

web/src/components/atoms/DownloadButton/index.tsx
nourbalaha: Done!

web/src/components/atoms/DownloadButton/index.tsx
nourbalaha: Done!

web/src/components/molecules/Asset/Asset/AssetBody/svgPreview.tsx
nourbalaha: Done!

web/src/components/molecules/Asset/AssetList/AssetListTable/index.tsx
nourbalaha: Done!
I added the `string | Date` type to the dateSortCallback function instead.

web/src/components/organisms/Asset/AssetList/assetSortType.type.ts
nourbalaha: Done!

issmail-basel (approved):
LGTM!

web/src/utils/uuidToURL.ts
```diff
@@ -0,0 +1,20 @@
+export const uuidToURL = (uuid: string, objectName: string) => {
+  // TODO: needs proper implementation
+  if (!uuid || !objectName) return "";
```
KaWaite, line 3: Since you typed the params to not allow them to be undefined, this isn't necessary, right?

web/src/utils/uuidToURL.ts
```diff
@@ -0,0 +1,20 @@
...
+  const storageHost = window.REEARTH_CONFIG?.storageHost ?? "http://localhost:8080";
```
KaWaite, line 4: The `storageHost` in the config currently if empty falls back to `"/"`. If it should be localhost:8080 (or "/" doesn't resolve to that), please set the env variable there. 
I think its better if you leave the env variables to the config and just if there is no storageHost return.

web/src/utils/uuidToURL.ts
```diff
@@ -0,0 +1,20 @@
...
+  const assetBase = window.REEARTH_CONFIG?.assetBase ?? "assets";
```
KaWaite, line 5: Same here

web/src/utils/uuidToURL.ts
```diff
@@ -0,0 +1,20 @@
...
+  const path = getPathFromUUID(uuid, objectName);
+  if (!path) return "";
+  return joinPath(storageHost, assetBase, path);
+};
+
+export const getPathFromUUID = (uuid: string, objectName: string) => {
+  if (!uuid || !objectName) return "";
```
KaWaite, line 12: Same as above. I dont htink this is necessary

web/src/components/organisms/Asset/convertAsset.ts
```diff
@@ -0,0 +1,17 @@
...
+    fileName: GQLAsset?.fileName ?? "",
+    createdAt: GQLAsset?.createdAt.toString() ?? "",
+    createdBy: GQLAsset?.createdBy?.name ?? "",
+    file: GQLAsset?.file as AssetFile,
+    previewType: GQLAsset?.previewType as PreviewType,
+    projectId: GQLAsset?.projectId ?? "",
+    size: GQLAsset?.size ?? 0,
+    url: GQLAsset?.uuid && GQLAsset?.fileName ? uuidToURL(GQLAsset?.uuid, GQLAsset?.fileName) : "",
+  };
+};
```
KaWaite, line 17: Isn't it better to return if no GQLAsset and then we can clean it up much more.
Below is just an example and probably has some issues. What do you think?
```
export const convertAsset = (GQLAsset: GQLAsset | undefined): Asset => {
if (!GQLAsset) return;
  return {
    id: GQLAsset.id ,
    fileName: GQLAsset.fileName,
    createdAt: GQLAsset.createdAt.toString(),
    createdBy: GQLAsset.createdBy?.name ,
    file: GQLAsset.file as AssetFile,
    previewType: GQLAsset.previewType as PreviewType,
    projectId: GQLAsset.projectId,
    size: GQLAsset.size,
    url: GQLAsset.uuid && GQLAsset.fileName ? uuidToURL(GQLAsset?.uuid, GQLAsset?.fileName) : "",
  };
};
```

web/src/components/organisms/Asset/AssetList/hooks.ts
```diff
@@ -74,7 +59,7 @@ export default (projectId?: string) => {
     variables: {
       projectId: projectId ?? "",
       pagination: pagination(sort),
-      sort: toGQLEnum(sort?.type),
+      sort: sort?.type as GQLSortType,
```
KaWaite, line 62: I think the original code, with the toGQLEnum is copied from reearth-web. I forget why we needed the toGQLEnum and all its complexity, but have you tested this thoroughly?

web/src/components/atoms/DownloadButton/index.tsx
```diff
@@ -18,8 +18,15 @@ const DownloadButton: React.FC<DownloadButtonProps> = ({
   displayDefaultIcon,
   ...props
 }) => {
-  const handleDownload: MouseEventHandler<HTMLElement> | undefined = () => {
-    fileDownload(url, filename);
+  const handleDownload: MouseEventHandler<HTMLElement> | undefined = async () => {
```
KaWaite, line 21: Please start using useCallback on basically all functions inside a component. A while ago we discussed with @rot1024 and the conclusion was basically "why not? if we do then we know the functions won't rerender needlessly".

web/src/components/molecules/Asset/Asset/AssetBody/previewTypeSelect.tsx
```diff
@@ -20,14 +15,14 @@ type Props = {
 
 export const PreviewTypeSelect: React.FC<Props> = ({ onTypeChange, style, value }) => {
   const previewTypeList = [
-    { id: 1, name: "PNG/JPEG/TIFF/SVG", value: PreviewType.Image },
+    { id: 1, name: "PNG/JPEG/TIFF/SVG", value: "IMAGE" },
```
KaWaite, line 18: Can you type this prevIewTypeList? value needs to be of PreviewType and I think its better to type that.
```
type PreviewTypeListItem = {
id: number,
name: string,
value PreviewItem
}
```

web/src/components/molecules/Asset/Asset/AssetBody/svgPreview.tsx
```diff
@@ -6,12 +6,15 @@ type Props = { url: string; svgRender: boolean };
...
-      });
+  const fetchData = useCallback(async () => {
+    const res = await fetch(url, {
+      method: "GET",
+      headers: {
+        "Access-Control-Allow-Origin": "*",
+      },
+    });
+    const text = await res.text();
+    setSvgText(text);
```
KaWaite, line 17: Does this not through an error?  Doesn't fetch return an object with status and everything that you should check if the status is 200 and all that? Its been a while since I've used fetch...

All that being said, is it actually necessary to go through the trouble of fetching this JUST for fallback text? Why do you want to display the svg text and not just a "Could not display svg"?

web/src/components/molecules/Asset/AssetList/AssetListTable/index.tsx
```diff
@@ -80,8 +79,8 @@ const AssetListTable: React.FC<AssetListTableProps> = ({
     },
     {
       title: "Created By",
-      dataIndex: "createdById",
-      key: "createdById",
+      dataIndex: "createdBy",
+      key: "createdBy",
```
KaWaite, line 83: Sorry nothing about this point. But in the render for Action please fix the DownloadButton. Its doing the <DownloadButton></DownloadButton> thing

KaWaite (commented):
Reviewed!

web/src/utils/uuidToURL.ts
nourbalaha: Done!
You're right.

web/src/utils/uuidToURL.ts
nourbalaha: Done!

web/src/utils/uuidToURL.ts
nourbalaha: Done!

web/src/utils/uuidToURL.ts
nourbalaha: Done!

web/src/components/organisms/Asset/convertAsset.ts
nourbalaha: Then the return type should be `Asset | undefined` instead of `Asset` and that type change should be done in multiple places. does it worth it?

web/src/components/atoms/DownloadButton/index.tsx
nourbalaha: Done!

web/src/components/molecules/Asset/Asset/AssetBody/previewTypeSelect.tsx
nourbalaha: Done!

web/src/components/molecules/Asset/AssetList/AssetListTable/index.tsx
nourbalaha: Can you explain more please?

web/src/components/organisms/Asset/AssetList/hooks.ts
nourbalaha: I haven't fully implemented sort, but it works for this PR.

web/src/components/molecules/Asset/Asset/AssetBody/svgPreview.tsx
nourbalaha: Do you mean "throw" an error? 
Do you mean that I should add an if statement to check the response.status?
however I added this line of code just in case:
    ```
    if (res.status !== 200) {
      setSvgText("Could not display svg");
      return;
    }
    ```

web/src/components/organisms/Asset/convertAsset.ts
KaWaite: Depends on how it's being used. 
But I'm guessing you are converting queried GQLAssets to Assets that will be displayed in some way (or at least affects the UI in some way). If that is the case, isn't it better to not display an empty asset/component?

web/src/components/molecules/Asset/Asset/AssetBody/svgPreview.tsx
KaWaite: That I think is better, thanks!

web/src/components/molecules/Asset/AssetList/AssetListTable/index.tsx
KaWaite: Sorry not sure what happened to my comment... 
What I meant is in this component's render you are using the DownloadButton component, right? Its written like <DownloadButton></DownloadButton>, which we want to avoid

nourbalaha: Done!

web/src/components/organisms/Asset/convertAsset.ts
nourbalaha: Done!

web/src/components/molecules/Asset/Asset/AssetBody/svgPreview.tsx
```diff
@@ -6,12 +6,19 @@ type Props = { url: string; svgRender: boolean };
...
-      });
+  const fetchData = useCallback(async () => {
+    const res = await fetch(url, {
+      method: "GET",
+      headers: {
+        "Access-Control-Allow-Origin": "*",
+      },
+    });
+    if (res.status !== 200) {
+      setSvgText("Could not display svg");
```
KaWaite, line 17: Sorry just remembered this needs to be i18n supported

nourbalaha: I will add i18n to all texts in Asset and AssetList pages in a new separate PR.

web/src/components/organisms/Asset/Asset/hooks.ts
```diff
@@ -1,31 +1,37 @@
...
+  const { data: rawAsset, loading } = useGetAssetQuery({
     variables: {
       assetId: assetId ?? "",
     },
   });
 
+  const asset: Asset | undefined = useMemo(() => {
+    return convertAsset(rawAsset?.asset as GQLAsset);
+  }, [rawAsset]);
+
```
KaWaite, line 27: If you want to display ONLY the assets that are available, and don't need to/want to display the "Asset not found" fallback, do like below and then you don't have to type like `assets?: Assets | undefined[]`. 
If you go this way, please test this and then update all types you just updated to not allow for undefined inside the assets array!

```suggestion
  const asset: Asset | undefined = useMemo(() => {
    return convertAsset(rawAsset?.asset as GQLAsset).filter((a): a is Asset => !!a);
  }, [rawAsset]);

```

web/src/components/molecules/Asset/Asset/AssetBody/svgPreview.tsx
KaWaite: Oh okay, I guess the i18n PR merged after you began the asset tasks. Gotcha

web/src/components/organisms/Asset/Asset/hooks.ts
nourbalaha: convertAsset function returns an `Asset` not an array of assets `Asset[]` so we can't use `.filter` here.
I think you meant that I should add `.filter` in `web/src/components/organisms/Asset/AssetList/hooks.ts` not in `web/src/components/organisms/Asset/Asset/hooks.ts`
something like this right?
```
  useEffect(() => {
    const assets =
      data?.assets.nodes
        .map(asset => asset as GQLAsset)
        .map(convertAsset)
        .filter(asset => !!asset) ?? [];
    setAssetList(assets as Asset[]);
  }, [data?.assets.nodes]);
```

web/src/components/organisms/Asset/AssetList/hooks.ts
```diff
@@ -181,8 +166,12 @@ export default (projectId?: string) => {
...
 
   useEffect(() => {
-    const assets = (data?.assets.nodes as AssetNode[]) ?? [];
-    setAssetList(assets);
+    const assets =
+      data?.assets.nodes
+        .map(asset => asset as GQLAsset)
+        .map(convertAsset)
+        .filter(asset => !!asset) ?? [];
+    setAssetList(assets as Asset[]);
```
KaWaite, line 174:
```suggestion
    const assets: Asset[] =
      data?.assets.nodes
        .map(asset => asset as GQLAsset)
        .map(convertAsset)
        .filter(asset => !!asset) ?? [];
    setAssetList(assets);
```

KaWaite: Where possible you should avoid type assertions!

nourbalaha: Although a filter is added I am still getting this error message:
```
Type '(Asset | undefined)[]' is not assignable to type 'Asset[]'.
  Type 'Asset | undefined' is not assignable to type 'Asset'.
    Type 'undefined' is not assignable to type 'Asset'.
```

KaWaite (approved):

PR #244, opened 2022-09-21
fix(web): content height

issmail-basel:
# Overview
content can be more than 100%, also `height` property causes an unintentional scroll.


nourbalaha (approved):
Approved!

KaWaite (approved):

PR #258, opened 2022-09-30
chore(web): open browser window when dev server runs

HideBa:
# Why
Because it's more convenient to open browser when web server runs


KaWaite (approved):
If you are sure you want this then merge away :P

PR #259, opened 2022-10-02
feat(web): integration hooks

issmail-basel:


web/src/components/molecules/Integration/types.ts
```diff
@@ -0,0 +1,51 @@
...
+export type Developer = {
+  id: string;
+  name: string;
+  email: string;
+};
+
+export enum IntegrationType {
+  Private = "Private",
+  Public = "Public",
+}
```
nourbalaha, line 33:
```suggestion
export type IntegrationType = "Private" | "Public"
```
maybe ?
If you do this change please don't forget to replace
`IntegrationType.Private` with `"Private"` and 
`IntegrationType.Public` with `"Public"`

nourbalaha (approved):
Approved with one suggestion!

web/src/components/molecules/Integration/types.ts
issmail-basel: I agree :+1:

web/src/components/molecules/Dashboard/types.ts
```diff
@@ -10,9 +10,11 @@ export type User = {
 
 export type Member = {
   userId: string;
+  role: any;
```
KaWaite, line 13: Can you not type this better?

web/src/gql/queries/workspace.ts
```diff
@@ -64,7 +64,46 @@ export const UPDATE_MEMBER_OF_WORKSPACE = gql`
 
 export const REMOVE_MEMBER_FROM_WORKSPACE = gql`
   mutation RemoveMemberFromWorkspace($workspaceId: ID!, $userId: ID!) {
-    removeMemberFromWorkspace(input: { workspaceId: $workspaceId, userId: $userId }) {
+    removeUserFromWorkspace(input: { workspaceId: $workspaceId, userId: $userId }) {
```
KaWaite, line 67: Can we keep language consistent with the backend by updating the mutation name as well as the variable name to be User instead of Member?

issmail-basel: @KaWaite , this is not a part of the PR but I did these temporary changes to not break the system.
We had one type of member and then changed it to user/integration on the backend.
Old members are now users.
The whole member section needs refactoring (queries/hooks) because of the backend edit, and it's not part of the integration PR.

web/src/components/molecules/Integration/IntegrationSettingsModal/index.tsx
```diff
@@ -1,39 +1,62 @@
...
+  }, [form, onClose, onSubmit]);
+
   return (
     <Modal
-      title={t("Integration Setting")}
+      title={t("Integration Setting") + "  " + selectedIntegrationMember?.integration?.name}
       visible={open}
       onCancel={() => onClose?.()}
-      onOk={onSubmit}
+      onOk={handleSubmit}
```
KaWaite, line 47: It doesn't seem like you are making use of the confirm functionality inside antd's modal, are you?
If not, the onOk isn't necessary since you are doing handleSubmit on the Save button click.

web/src/components/molecules/Integration/IntegrationTable/index.tsx
```diff
@@ -3,75 +3,125 @@ import styled from "@emotion/styled";
...
   const handleToolbarEvents: ListToolBarProps | undefined = {
     search: {
-      onSearch: _ => {},
+      onSearch: (value: string) => {
+        if (value) {
+          onSearchTerm(value);
+        } else {
+          onSearchTerm();
+        }
+      },
```
KaWaite, line 75: value is of type string so it'll never reach the else statement.

web/src/components/molecules/Integration/IntegrationTable/index.tsx
```diff
@@ -3,75 +3,125 @@ import styled from "@emotion/styled";
...
+                onClick={onIntegrationConnectModalOpen}
+                type="primary"
+                icon={<Icon icon="api" />}>
+                {t("Connect Integration")}
+              </Button>
+            </Suggestion>
+            <Suggestion>
+              {t("Or read")} <a href="">{t("how to use Re:Earth CMS")}</a> {t("first")}
+            </Suggestion>
+          </EmptyTableWrapper>
```
KaWaite, line 105: I guess it was in the other PR that we discussed this being different than the design. Please update here too..

web/src/components/organisms/Settings/Integration/hooks.ts
```diff
@@ -0,0 +1,152 @@
...
+export default (workspaceId?: string) => {
+  const [selectedConnectionModalIntegration, SetSelectedConnectionModalIntegration] =
+    useState<Integration>();
+  const [selectedIntegrationMember, SetSelectedIntegrationMember] = useState<IntegrationMember>();
+  const [integrationConnectModalShown, setIntegrationConnectModalShown] = useState(false);
+  const [integrationSettingsModalShown, setIntegrationSettingsModalShown] = useState(false);
+  const [searchTerm, setSearchTerm] = useState<string>();
+  const { data, refetch } = useGetMeQuery();
+
+  const workspaces = data?.me?.workspaces;
```
KaWaite, line 24: useMemo

web/src/components/organisms/Settings/Integration/hooks.ts
```diff
@@ -0,0 +1,152 @@
...
+  const workspace = workspaces?.find(workspace => workspace.id === workspaceId);
+
+  const fromIntegration = (integration: any) => ({
```
KaWaite, line 27: `integration: any` should be avoided

web/src/components/organisms/Settings/Integration/hooks.ts
```diff
@@ -0,0 +1,152 @@
...
+    config: {
+      token: integration.config?.token,
+      webhooks: integration.config?.webhooks,
+    },
+  });
+
+  const integrations = useMemo(() => {
+    return (data?.me?.integrations ?? [])
+      .map<Integration | undefined>(integration =>
+        integration ? fromIntegration(integration) : undefined,
```
KaWaite, line 44: Make the `fromIntegration` function return undefined if integration is undefined and then you don't need the ternary operator here.

web/src/components/organisms/Settings/Integration/hooks.ts
```diff
@@ -0,0 +1,152 @@
...
+      )
+      .filter((integration): integration is Integration => !!integration);
+  }, [data?.me?.integrations]);
+
+  const workspaceIntegrationMembers = useMemo(() => {
+    return (workspace?.members ?? [])
+      .map<IntegrationMember | undefined>(member =>
```
KaWaite, line 51: Again, please update all places that are using this style to the decided upon style (blah?.blah?.map())

web/src/components/organisms/Settings/Integration/hooks.ts
```diff
@@ -0,0 +1,152 @@
...
+} from "@reearth-cms/components/molecules/Integration/types";
+import {
+  useGetMeQuery,
+  useAddIntegrationToWorkspaceMutation,
+  Role as GQLRole,
+  useUpdateIntegrationOfWorkspaceMutation,
+} from "@reearth-cms/gql/graphql-client-api";
+
+export default (workspaceId?: string) => {
+  const [selectedConnectionModalIntegration, SetSelectedConnectionModalIntegration] =
```
KaWaite, line 16: This name is confusing. Is it just a selectedIntegration? Or is it actually "an integration that is a connection modal that is selected"??

web/src/components/organisms/Settings/Integration/hooks.ts
```diff
@@ -0,0 +1,152 @@
...
+    SetSelectedConnectionModalIntegration(undefined);
+    setIntegrationConnectModalShown(true);
+  }, []);
+
+  const handleIntegrationSettingsModalClose = useCallback(() => {
+    setIntegrationSettingsModalShown(false);
+  }, []);
+
+  const handleIntegrationSettingsModalOpen = useCallback((integrationMember: IntegrationMember) => {
+    console.log(integrationMember);
```
KaWaite, line 83: remove please

web/src/components/organisms/Settings/Integration/hooks.ts
```diff
@@ -0,0 +1,152 @@
...
+      variables: {
+        integrationId: selectedConnectionModalIntegration.id,
+        workspaceId,
+        role: GQLRole.Reader,
+      },
+    });
+    if (integration.errors || !integration.data?.addIntegrationToWorkspace) {
+      setIntegrationConnectModalShown(false);
+    }
+    setIntegrationConnectModalShown(false);
```
KaWaite, line 107: The error if statement is redundant

web/src/components/organisms/Settings/Integration/hooks.ts
```diff
@@ -0,0 +1,152 @@
...
+        variables: {
+          integrationId: selectedIntegrationMember?.integration.id,
+          workspaceId,
+          role: role as GQLRole,
+        },
+      });
+      if (integration.errors || !integration.data?.updateIntegrationOfWorkspace) {
+        setIntegrationConnectModalShown(false);
+      }
+      setIntegrationConnectModalShown(false);
```
KaWaite, line 126: Again, redundant. Please be careful when copy & pasting that the original code is good. Please update anywhere else its like this

web/src/components/organisms/Settings/Members/hooks.ts
```diff
@@ -59,6 +59,20 @@ export default ({ workspaceId }: Props) => {
     [searchUserQuery],
   );
 
+  const workspaceUserMembers = useMemo((): Member[] => {
+    return (currentWorkspace?.members ?? [])
```
KaWaite, line 63: here too

KaWaite (commented):
Reviewed

web/src/components/molecules/Integration/IntegrationTable/index.tsx
issmail-basel: @KaWaite , this one is different, the other one was related to adding items, and it didn't have this action.
But for integrations, it has:

web/src/components/organisms/Settings/Integration/hooks.ts
issmail-basel: @KaWaite , yes it is related only to the modal, because there is an `IntegrationMember` selected as well.
To avoid confusion I named it like this.

web/src/components/molecules/Integration/IntegrationSettingsModal/index.tsx
issmail-basel: I'm not using it in this modal, ok :+1:

web/src/components/molecules/Integration/IntegrationTable/index.tsx
issmail-basel: @KaWaite , my fault I have copied this from the Assets search without much thinking about the statement.
Fixed, and tested as well.

issmail-basel: https://github.com/reearth/reearth-cms/blob/58c8be6313ec45a489b63534dab11973f18ce352/web/src/components/molecules/Integration/IntegrationTable/index.tsx#L69-L71

web/src/components/organisms/Settings/Integration/hooks.ts
issmail-basel: https://github.com/reearth/reearth-cms/blob/58c8be6313ec45a489b63534dab11973f18ce352/web/src/components/organisms/Settings/Integration/hooks.ts#L25

web/src/components/organisms/Settings/Integration/hooks.ts
issmail-basel: https://github.com/reearth/reearth-cms/blob/58c8be6313ec45a489b63534dab11973f18ce352/web/src/components/organisms/Settings/Integration/hooks.ts#L28

web/src/components/organisms/Settings/Integration/hooks.ts
issmail-basel: Done.

web/src/components/organisms/Settings/Integration/hooks.ts
issmail-basel: https://github.com/reearth/reearth-cms/blob/58c8be6313ec45a489b63534dab11973f18ce352/web/src/components/organisms/Settings/Integration/hooks.ts#L42-L46

web/src/components/organisms/Settings/Integration/hooks.ts
issmail-basel: https://github.com/reearth/reearth-cms/blob/58c8be6313ec45a489b63534dab11973f18ce352/web/src/components/organisms/Settings/Integration/hooks.ts#L49-L51

web/src/components/organisms/Settings/Integration/hooks.ts
issmail-basel: Removed.

web/src/components/organisms/Settings/Members/hooks.ts
issmail-basel: Done

web/src/components/organisms/Settings/Integration/hooks.ts
issmail-basel: https://github.com/reearth/reearth-cms/blob/58c8be6313ec45a489b63534dab11973f18ce352/web/src/components/organisms/Settings/Integration/hooks.ts#L121-L127

web/src/components/organisms/Settings/Integration/hooks.ts
issmail-basel: Added comment to add notification later, and prevented modal from closing in case of error (like the other one below).

web/src/components/molecules/Integration/IntegrationTable/index.tsx
KaWaite: @lavalse Do you want them different like that?

web/src/components/organisms/Settings/Integration/hooks.ts
KaWaite: @issmail-basel Is there any way you can move this inside of the modal itself, it can handle it and keep things simpler? If you move this in to the modal, you can have a selectedIntegration prop that you pass whichever integration has been selected for modal open. Then, on close just clear this field (with any other fields you need to clear).
What do you think?

web/src/components/organisms/Settings/Members/hooks.ts
```diff
@@ -59,6 +59,20 @@ export default ({ workspaceId }: Props) => {
     [searchUserQuery],
   );
 
+  const workspaceUserMembers = useMemo((): Member[] | undefined => {
```
KaWaite, line 62: Sorry I thought I mentioned this in the first review but must have forgot to comment about this. Shouldn't this be either workspaceUsers OR workspaceMembers? Why workspaceUserMembers?

issmail-basel: @KaWaite , at first backend only had **members** in each workspace, then member became one of two possible types: either **userMembers** or **integrationMembers**.

web/src/components/organisms/Settings/Integration/hooks.ts
issmail-basel: @KaWaite , moved the state inside the modal itself, the code is more readable now in my opinion.

https://github.com/reearth/reearth-cms/blob/0d2d4093be036c6fad8fb94a23aabf14657cf87e/web/src/components/molecules/Integration/IntegrationConnectModal/index.tsx#L19-L27

web/src/components/organisms/Settings/Members/hooks.ts
issmail-basel: @KaWaite , btw, I'll create another PR, where all members (old **Member**) become **UserMembers**.
The main focus here is integration right now.

web/src/components/molecules/Integration/IntegrationTable/index.tsx
KaWaite: So this is fine for now.

KaWaite (approved):

PR #261, opened 2022-10-05
feat(server): item search

mimoham24:
# Overview


server/schemas/item.graphql
```diff
@@ -71,6 +71,13 @@ extend type Query {
     before: Cursor
   ): ItemConnection!
   versionsByItem(itemId: ID!): [VersionedItem!]!
+  searchItem(
+    key: String!
```
rot1024, line 75: Given that we will extend more advanced queries in the future, the current input is rough. Let's define the query in a bit more detail.

```gql
input ItemQuery {
  workspace: ID!
  project: ID!
  q: String
  # add more queries here in the future
}
```

Then

```suggestion
    key: ItemQuery!
```

Also at `pkg/item`, define domain model:

```go
type Query struct {
  workspace id.WorkspaceID
  project id.ProjectID
  q string
}

func NewQuery(workpsace id.ProjectID, project *id.ProjectID, q string) *Query {
	return &Query{
		workpsace: workspace,
		project: project,
		q: q,
	}
}

// Q returns keywords for search
func (q *Query) Q() string {
	return q.q
}

// define Workspace and Project method also
```

Then use it in `repo/Item`:

```go
type Item interface {
  Search(context.Context, *item.Query, *usecasex.Pagination) (item.List, *usecasex.PageInfo, error)
}
```

rot1024 (changes requested):

server/pkg/version/value.go
```diff
@@ -1,5 +1,7 @@
 package version
 
+import "fmt"
```
rot1024: not used

server/internal/adapter/gql/loader_item.go
```diff
@@ -114,8 +114,17 @@ func (c *ItemLoader) FindByProject(ctx context.Context, projectID gqlmodel.ID, f
...
+func (c *ItemLoader) Search(ctx context.Context, query gqlmodel.ItemQuery, first *int, last *int, before *usecasex.Cursor, after *usecasex.Cursor) (*gqlmodel.ItemConnection, error) {
+	wid, err := gqlmodel.ToID[id.Workspace](query.Workspace)
+	if err != nil {
+		return nil, err
+	}
+	pid, err := gqlmodel.ToID[id.Project](query.Project)
+	if err != nil {
+		return nil, err
+	}
+	q := item.NewQuery(wid, pid, *query.Q)
```
rot1024, line 126: Defining `func ToItemQuery(ItemQuery) *item.Query` is good

rot1024 (changes requested):

rot1024 (approved):

PR #266, opened 2022-10-06
feat(web): add content table hook

issmail-basel:
# Overview
Link content table to items backend


web/src/components/molecules/Content/ContentTable/index.tsx
```diff
@@ -1,10 +1,12 @@
 import ProTable, { ListToolBarProps, ProColumns } from "@reearth-cms/components/atoms/ProTable";
+import { Item } from "@reearth-cms/components/molecules/Content/types";
 
-const columns: ProColumns<any>[] = [];
-
-const ContentTable: React.FC = () => {
-  const dataSource: [] = [];
+export type Props = {
+  contentTableFields?: any[];
```
KaWaite, line 5: As always, can't you type this better than `any`?

web/src/components/organisms/Project/Content/ContentList/index.tsx
```diff
@@ -25,7 +29,11 @@ const ContentList: React.FC = () => {
     <PaddedContent>
       <StyledModelsMenu title={t("Content")} selectModel={selectModel} />
       <ContentChild>
-        <ContentTable />
+        <ContentTableWrapper
```
KaWaite, line 32: Can you move this Wrapper's parent `ContentChild`'s styles in to the ContentTableWrapper? Currently this organism has styles (which it shouldn't as we discussed) and it makes more sense that the ContentTableWRAPPER has the wrapper that is wrapping its content :P

web/src/components/organisms/Project/Content/ContentList/hooks.ts
```diff
@@ -0,0 +1,44 @@
...
+                (obj, field) => Object.assign(obj, { [field.schemaFieldID]: field.value }),
+                {},
+              ),
+            }
+          : undefined,
+      )
+      .filter((contentTableField): contentTableField is ContentTableField => !!contentTableField);
+  }, [data?.items.nodes]);
+
+  const contentTableColumns = useMemo((): ProColumns<Item>[] | undefined => {
```
KaWaite, line 31: Can you move the typings to the variable, not the useMemo callback? I feel like it'd be clearer that way. Also to the contentTableFields above!

web/src/components/molecules/Content/types.ts
```diff
@@ -0,0 +1,19 @@
...
+export type Item = {
+  id: string;
+  schemaID: string;
+  fields: ItemField[] | undefined | null;
+};
+
+export type ContentTableField = {
+  id: string;
+  schemaID: string;
+  fields: { [key: string]: any };
```
KaWaite, line 18: Can the value actually be anything? Or can you type this better?

web/src/components/molecules/Content/types.ts
```diff
@@ -0,0 +1,19 @@
+import { FieldType } from "@reearth-cms/components/molecules/Schema/types";
+
+export type ItemField = {
+  schemaFieldID: string;
+  type: FieldType;
+  value: any;
```
KaWaite, line 6: Can you type this better?

KaWaite (commented):
Reviewed

web/src/components/molecules/Content/ContentTable/index.tsx
issmail-basel: @KaWaite , Backend typed it as `any`, it can be any type from string to an integer to array...etc

web/src/components/molecules/Content/types.ts
issmail-basel: @KaWaite , yes it can be anything

web/src/components/molecules/Content/types.ts
issmail-basel: Same as above, value can be anything.

issmail-basel: @KaWaite , Backend typed it as `any`, it can be any type from string to an integer to array...etc

web/src/components/molecules/Content/types.ts
KaWaite: ok!

web/src/components/molecules/Content/ContentTable/index.tsx
issmail-basel: Fixed in all places now.

web/src/components/organisms/Project/Content/ContentList/index.tsx
issmail-basel: https://github.com/reearth/reearth-cms/blob/25bf6fb1124a2ffa260fcd8fb8de42741f96f325/web/src/components/organisms/Project/Content/ContentList/index.tsx#L26-L32

KaWaite (approved):

Issue #28, opened 2022-03-24
Public API of the project


HideBa:
# Leader
@

# User story
Users want to access their own data through public APIs. This will allow them to edit their data in the CMS and use it via the API for use in blogs, Re:Earth itself, and other applications. Also, by specifying the data format using the extension at the end of the endpoint, the data can be returned in a format that is supported by Re:Earth by default and can be used by other applications such as blogs, Re:Earth itself, and other applications.

# Requirements
- [ ] Tables of the project whose publication settings allows to them be public can be retrieved via public API
- [ ] Re:Earth should support this data format as default
	- [ ] `tables/tableId/records.json` -> JSON
	- [ ] `tables/tableId/records.csv` -> CSV

# Design file

# Testing list
- [ ] When OO is
   - [ ] If you input "XXX", it should show "YYY"

PR #267, opened 2022-10-06
feat(server): Implement from domain models to Adapter on the public

gop50k:
# Overview
Implementation of Public APIs
From domain models to Adapter, implemented in project and model.

# Ticket
https://github.com/reearth/reearth-cms/issues/28


server/internal/adapter/gql/resolver_mutation_model.go
```diff
@@ -13,12 +13,13 @@ func (r *mutationResolver) CreateModel(ctx context.Context, input gqlmodel.Creat
...
 		return nil, err
 	}
-
+	var m *interfaces.CreateModelParam
 	res, err := usecases(ctx).Model.Create(ctx, interfaces.CreateModelParam{
 		ProjectId:   pId,
 		Name:        input.Name,
 		Description: input.Description,
 		Key:         input.Key,
+		Public:      m.Public,
```
rot1024, line 22:
```suggestion
		Public:      false,
```

server/internal/adapter/gql/resolver_mutation_model.go
```diff
@@ -35,11 +36,13 @@ func (r *mutationResolver) UpdateModel(ctx context.Context, input gqlmodel.Updat
 		return nil, err
 	}
 
+	var m *interfaces.UpdateModelParam
 	res, err := usecases(ctx).Model.Update(ctx, interfaces.UpdateModelParam{
 		ModelId:     mId,
 		Name:        input.Name,
 		Description: input.Description,
 		Key:         input.Key,
+		Public:      m.Public,
```
rot1024, line 45:
```suggestion
		Public:      input.Public,
```

server/internal/usecase/interactor/model.go
```diff
@@ -80,6 +80,9 @@ func (i Model) Create(ctx context.Context, param interfaces.CreateModelParam, op
 			if param.Description != nil {
 				mb = mb.Description(*param.Description)
 			}
+			if param.Public != nil {
+				mb = mb.IsPublic(*param.Public)
```
rot1024, line 84: Rename IsPablic to Public

server/pkg/model/model_test.go
```diff
@@ -214,12 +214,19 @@ func TestModel_Public(t *testing.T) {
 		want  bool
 	}{
 		{
-			name: "test",
+			name: "set public true",
```
rot1024, line 217:
```suggestion
			name: "public true",
```

Confusing name; nothing is set.

server/pkg/model/model_test.go
```diff
@@ -214,12 +214,19 @@ func TestModel_Public(t *testing.T) {
...
 			model: Model{
 				public: true,
 			},
 			want: true,
 		},
+		{
+			name: "set public false",
```
rot1024, line 224:
```suggestion
			name: "public false",
```

server/pkg/model/model_test.go
```diff
@@ -374,7 +381,7 @@ func TestModel_SetPublic(t *testing.T) {
 		args args
 	}{
 		{
-			name: "test",
+			name: "set public true",
 			args: args{
 				public: true,
```
rot1024, line 386: public: false

We want to test if the field is correctly changed:

True -> false or false -> true

server/pkg/model/model_test.go
```diff
@@ -383,7 +390,7 @@ func TestModel_SetPublic(t *testing.T) {
 			},
 		},
 		{
-			name: "test",
+			name: "set public faalse",
 			args: args{
 				public: false,
```
rot1024, line 395: public: true

server/internal/adapter/gql/resolver_mutation_model.go
```diff
@@ -13,12 +13,13 @@ func (r *mutationResolver) CreateModel(ctx context.Context, input gqlmodel.Creat
 	if err != nil {
 		return nil, err
 	}
-
+	var m *interfaces.CreateModelParam
```
rot1024, line 16: Can you please explain why this is necessary? Seems like an unnecessary change to me.

rot1024 (changes requested):

server/internal/adapter/gql/resolver_mutation_model.go
gop50k: @rot1024 
I thought it was not a good idea to add a Public field to an already generated CreateModelParam, so I set it as a variable in advance.

```
input is defined like this
input gqlmodel.CreateModelInput
type CreateModelInput struct {
	ProjectID   ID      `json:"projectId"`
	Name        *string `json:"name"`
	Description *string `json:"description"`
	Key         *string `json:"key"`
}


input gqlmodel.UpdateModelInput
type UpdateModelInput struct {
	ModelID     ID      `json:"modelId"`
	Name        *string `json:"name"`
	Description *string `json:"description"`
	Key         *string `json:"key"`
}
```

server/internal/adapter/gql/resolver_mutation_model.go
```diff
@@ -6,19 +6,20 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
-
 	res, err := usecases(ctx).Model.Create(ctx, interfaces.CreateModelParam{
 		ProjectId:   pId,
 		Name:        input.Name,
 		Description: input.Description,
 		Key:         input.Key,
+		Public:      lo.ToPtr(false),
```
rot1024, line 22:
```suggestion
		Public:      nil,
```

nil is ok

rot1024 (approved):

PR #270, opened 2022-10-07
feat(web): add content form

issmail-basel:
# Overview


web/src/components/molecules/Content/ContentDetailsMolecule/index.tsx
```diff
@@ -0,0 +1,64 @@
+import styled from "@emotion/styled";
```
KaWaite, line 1: Please rename this component's folder to ContentDetails. its a child of the molecules directory so we don't need this to be ContentDetailsMolecule

web/src/components/molecules/Content/ContentForm/index.tsx
```diff
@@ -0,0 +1,181 @@
+import styled from "@emotion/styled";
+import Upload from "antd/lib/upload/Upload";
```
KaWaite, line 2: Don't use directly

web/src/components/molecules/Content/ContentForm/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+    } catch (info) {
+      console.log("Validate Failed:", info);
+    }
+  }, [form, model?.schema.fields, model?.schema.id, itemID, onItemCreate, onItemUpdate]);
+
+  return (
+    <Form
+      form={form}
+      layout="vertical"
+      initialValues={itemID ? initialFormValues : defaultFormValues}>
```
KaWaite, line 64: This has the potential to be undefined (because you allow initialFormValues to be `any`), which since you have default values I'm guessing you don't want.

web/src/components/molecules/Content/ContentDetailsMolecule/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+          onBack={onBack}
+          itemId={itemId}
+          onItemCreate={onItemCreate}
+          onItemUpdate={onItemUpdate}
+          model={model}
+          defaultFormValues={defaultFormValues}
+          initialFormValues={initialFormValues}
+        />
+      </ContentChild>
+    </PaddedContent>
```
KaWaite, line 43: You repeat this style in many places. Please make a generic component and use that instead of copy/pasting a bunch.

web/src/components/molecules/Content/ContentTable/index.tsx
```diff
@@ -13,15 +16,25 @@ const ContentTable: React.FC<Props> = ({ contentTableFields, contentTableColumns
...
-      columns={contentTableColumns}
+      columns={[
+        {
+          title: "actions",
+          render: (_, contentField) => (
+            <Link to={"./details/" + contentField.id}>
+              <Icon icon="edit" />
+            </Link>
+          ),
+        },
```
KaWaite, line 30: I tried looking but couldn't figure it out. Why are you not including this column with the `contentTableColumns`?

web/src/components/organisms/Project/Content/ContentDetails/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const t = useT();
+
+  const navigate = useNavigate();
+
+  const selectModel = useCallback(
+    (modelId: string) => {
+      navigate(`/workspaces/${workspaceId}/${projectId}/content/${modelId}`);
+    },
+    [navigate, workspaceId, projectId],
+  );
```
KaWaite, line 23: This navigates somewhere. It doesn't select a model. Similar to the callback below, perhaps handleNavigateToModel or something is clearer

web/src/components/organisms/Project/Content/ContentDetails/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  const handleNvigateBack = useCallback(() => {
```
KaWaite, line 25: const handleNavigateBack = useCallback(() => {

web/src/components/organisms/Project/Content/ContentList/index.tsx
```diff
@@ -23,9 +23,14 @@ const ContentList: React.FC = () => {
     [navigate, workspaceId, projectId],
   );
 
+  const handleItemAdd = useCallback(() => {
+    navigate(`/workspaces/${workspaceId}/${projectId}/content/${modelId}/details`);
+  }, [navigate, workspaceId, projectId, modelId]);
+
```
KaWaite, line 29: This doesn't handle adding an item. Its a redirect. Same as above, please rename this more appropriately

web/src/components/organisms/Project/Content/hooks.ts
```diff
@@ -0,0 +1,15 @@
+import { useGetItemsQuery } from "@reearth-cms/gql/graphql-client-api";
+import { useModel } from "@reearth-cms/state";
+
+export default () => {
```
KaWaite, line 4: Is this being used? Its a new file but the Content/index isn't using it?

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
```diff
@@ -0,0 +1,100 @@
...
+} from "@reearth-cms/gql/graphql-client-api";
+
+import useContentHooks from "../hooks";
+
+type Props = {
+  itemId: string | undefined;
+};
+
+export default ({ itemId: itemID }: Props) => {
+  const { currentModel, itemsData } = useContentHooks();
```
KaWaite, line 17: It looks like you are setting currentModel in Jotai state. Why don't you import in `useModel` here instead of using another component's hooks. 
As for itemsData, is it not desirable to hold that in Jotai state as well ?

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
```diff
@@ -0,0 +1,100 @@
...
+
+      const item = await createNewItem({
+        variables: {
+          schemaID: data.schemaID,
+          fields: data.fields.map(field => ({ ...field, type: field.type as SchemaFiledType })),
+        },
+      });
+      if (item.errors || !item.data?.createItem) {
+        return;
+      }
```
KaWaite, line 38: No need to return. Please add a comment for error handling or add notifications or whatever instead of the return

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
```diff
@@ -0,0 +1,100 @@
...
+  const [createNewItem] = useCreateItemMutation({
+    refetchQueries: ["GetItems"],
+  });
+
+  const handleItemCreate = useCallback(
+    async (data: {
+      schemaID: string;
+      fields: { schemaFieldID: string; type: FieldType; value: string }[];
+    }) => {
+      if (!data.schemaID) return;
```
KaWaite, line 28: This is redundant because data AND schemaID are mandatory

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
```diff
@@ -0,0 +1,100 @@
...
+  const [updateItem] = useUpdateItemMutation({
+    refetchQueries: ["GetItems"],
+  });
+
+  const handleItemUpdate = useCallback(
+    async (data: {
+      itemID: string;
+      fields: { schemaFieldID: string; type: FieldType; value: string }[];
+    }) => {
+      if (!data.itemID) return;
```
KaWaite, line 52: Same here as above

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
```diff
@@ -0,0 +1,100 @@
...
+      const item = await updateItem({
+        variables: {
+          itemID: data.itemID,
+          fields: data.fields.map(field => ({ ...field, type: field.type as SchemaFiledType })),
+        },
+      });
+      if (item.errors || !item.data?.updateItem) {
+        return;
```
KaWaite, line 60: Same as above

web/src/components/molecules/Content/ContentDetailsMolecule/index.tsx
issmail-basel: Renamed the folder.

web/src/components/molecules/Content/ContentForm/index.tsx
issmail-basel: Fixed, I should not depend totally on autocomplete, didn't notice.

web/src/components/organisms/Project/Content/ContentDetails/index.tsx
issmail-basel: Renamed to `handleNavigateToModel`.

web/src/components/organisms/Project/Content/ContentDetails/index.tsx
issmail-basel: Fixed typo.

web/src/components/organisms/Project/Content/ContentList/index.tsx
issmail-basel: Renamed to `handleNavigateToItemForm`.

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
issmail-basel: `TODO: notification` Comment added.

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
issmail-basel: You are right, fixed.

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
issmail-basel: Removed the unnecessary condition.

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
issmail-basel: `TODO: notification` Comment added.

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
issmail-basel: @KaWaite , the **currentModel** is set in the `modelMenu` organism (hook) and components with no direct relation to the modelsMenu component use its state and a couple of components share the state.

On the other hand, `ItemsData` state is needed on a couple of screens which has a direct relationship with the items (content table and content form), and the two pages are using the same query, so query caching does the trick to not load the **graphQL** more than once.

issmail-basel: @KaWaite , the **currentModel** is set in the `modelMenu` organism (hook) and components with no direct relation to the modelsMenu component use its state and a couple of components share the state.

On the other hand, `ItemsData` state is needed on a couple of screens which has a direct relationship with the items (content table and content form), and the two pages are using the same query, so query caching does the trick to not load the **graphQL** more than once.

web/src/components/organisms/Project/Content/hooks.ts
issmail-basel: @KaWaite  It is shared and used by the two child hooks
`Content/ContentDetails/hooks`
`Content/ContentList/hooks`

web/src/components/molecules/Content/ContentTable/index.tsx
issmail-basel: @KaWaite , this added object is using: 
- an atom that is the `Icon`
- the render function is returning `JSX`

`contentTableColumns` is generated in the hooks file `src/components/organisms/Project/Content/ContentList/hooks.tsx`
https://github.com/reearth/reearth-cms/blob/0e3c61829cf9937d29e61d041f60b69d02d829ce/web/src/components/organisms/Project/Content/ContentList/hooks.ts#L27-L33

So I needed a way to add this first column to the columns, without violating the above 2 points.

web/src/components/molecules/Content/ContentForm/index.tsx
issmail-basel: @KaWaite , it is typed as `any` from the backend, in frontend it is looping through the backend data keys, so the only way to make it have an `undefined` value, is to have the backend return undefined for it.
Also if it is `undefined` or `null`, there won't be a value in form and that's the correct behavior.

web/src/components/molecules/Content/ContentDetailsMolecule/index.tsx
issmail-basel: @KaWaite  I made a generic component `ContentWrapper` which contain these parts (PaddedContent wrapping a LeftPaneWrapper, ContentChild).

web/src/components/molecules/Content/ContentDetailsMolecule/index.tsx
KaWaite: Sorry my bad, if you can take it one step further and rename the inner folders, removing the Content since they are already in the content folder...

web/src/components/molecules/Content/ContentForm/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+  initialFormValues: any;
+  defaultFormValues: any;
+  model?: Model;
+  onItemCreate: (data: { schemaID: string; fields: ItemField[] }) => Promise<void>;
+  onItemUpdate: (data: { itemID: string; fields: ItemField[] }) => Promise<void>;
+  onBack: () => void;
+}
+
+const ContentForm: React.FC<Props> = ({
+  itemId: itemID,
```
KaWaite, line 29: Why do you need to change the name? If possible, please keep it `itemId` and keep it consistent throughout the code.

web/src/components/molecules/Content/ContentForm/index.tsx
KaWaite: Ok that much makes sense, but the logic can be cleared up a little more.
So, you have an array of initial values, and an array of default values. 

Some thoughts:

When you generate the initial values, you are already filtering with `itemId`, so if there is no itemId, initialFormValues will become undefined. So `initialFormValues ?? defaultFormValues` would be enough in this situation.

Then again, when an item is created, is it fully blank, and only user changed fields will have values, just like reearth-web? If so, you'll need to combine the initial and default values (looping over the schema).

web/src/components/molecules/Content/ContentTable/index.tsx
KaWaite: It is illegal to use router in molecules though. Is there a reason you are using Link here when you seem to typically use navigate most other places? If you don't need Link for some specific reason, use navigate so you can create a callback that can be passed to the Icon. That way, you can keep atoms out of the organism and have your redirect

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
KaWaite: Ok, for now this is fine. But I do want to discuss with you and other FE members soon about "global" or "shared" hooks files. For a long time with reearth-web we have dealt with duplicate hooks and its really not nice.

web/src/components/molecules/Content/ContentDetailsMolecule/index.tsx
issmail-basel: Done, renamed all the folders.

web/src/components/molecules/Content/ContentForm/index.tsx
issmail-basel: @KaWaite , no initial value is only showing when creating new content.
initialFormValues returns an empty object, and I need to check if it's updating or creating, for that purpose I'm using: 
`itemID ? initialFormValues : defaultFormValues`

web/src/components/molecules/Content/ContentForm/index.tsx
issmail-basel: @KaWaite , so it will be easier for me to remember to refactor and to refactor when the backend is done fixing the variable name.
I will rename the variable when @mimoham24 PR is approved.

web/src/components/molecules/Content/ContentTable/index.tsx
issmail-basel: @KaWaite , done, replaced the link with navigate in the organism.

web/src/components/molecules/Content/ContentForm/index.tsx
KaWaite: Ah the backend is changing the name as well? Ok

KaWaite (approved):

PR #271, opened 2022-10-07
fix(server): fetching old members causing an error

yk-eukarya:
# Overview

consider the user invited by himself if the invitedby id is not exists


server/internal/infrastructure/mongo/mongodoc/workspace.go
```diff
@@ -60,10 +60,14 @@ func (d *WorkspaceDocument) Model() (*user.Workspace, error) {
 			if err != nil {
 				return nil, err
 			}
+			inviterId, err := id.UserIDFrom(member.InvitedBy)
```
rot1024, line 63:
```suggestion
			inviterID, err := id.UserIDFrom(member.InvitedBy)
```

rot1024 (approved):

PR #275, opened 2022-10-11
feat(web): Accessibility page

KaWaite:
# Overview


web/src/components/molecules/Accessibility/index.tsx
```diff
@@ -0,0 +1,154 @@
+import styled from "@emotion/styled";
+import { Switch } from "antd";
```
nourbalaha, line 2: Switch should be copied to atoms and imported from there, right?

nourbalaha (approved):
Approved with one requested change.

PR #276, opened 2022-10-11
feat(sever): add project ID to the item

mimoham24:
# Overview

## What I haven't done
I'll implement the project filter on another branch


server/pkg/item/item.go
```diff
@@ -22,6 +24,10 @@ func (i *Item) Fields() []*Field {
 	return slices.Clone(i.fields)
 }
 
+func (i *Item) Project() project.ID {
```
rot1024, line 27:
```suggestion
func (i *Item) Project() ProjectID {
```

server/pkg/item/item.go
```diff
@@ -9,9 +10,10 @@ import (
 type VersionedItem = version.Value[Item]
 
 type Item struct {
-	id       ID
-	schemaID schema.ID
-	fields   []*Field
+	id        ID
+	schemaID  schema.ID
+	projectID project.ID
```
rot1024, line 15:
```suggestion
	projectID ProjectID
```

Due to the danger of circular references, do not import a package if you will only use the ID of that package; import only the id package.

Define `type ProjectID = id.ProjectID` at `pkg/item/id.go`

server/schemas/item.graphql
```diff
@@ -26,6 +27,7 @@ input ItemFieldInput {
 
 input CreateItemInput {
   schemaId: ID!
+  projectId: ID!
```
rot1024, line 30:
```suggestion
```

That is not necessary since the schemaId alone identifies the project.

server/schemas/item.graphql
```diff
@@ -1,6 +1,7 @@
 type Item implements Node {
   id: ID!
   schemaId: ID!
+  projectId: ID!
   fields: [ItemField!]
```
rot1024, line 5:
```suggestion
  fields: [ItemField!]
  schema: Schema
  project: Project
```

Add resolvers for convenience.

server/schemas/item.graphql
```diff
@@ -67,6 +69,13 @@ extend type Query {
     after: Cursor
     before: Cursor
   ): ItemConnection!
+  itemsByProject(
+    projectId: ID!
+    first: Int
+    last: Int
+    after: Cursor
+    before: Cursor
+  ): ItemConnection!
```
rot1024, line 78:
```suggestion
```

It is not very useful to get all items across the schema directly from the project.

rot1024 (changes requested):
Create indexes for the Item collection. Currenlty there are no indexes and it causes serious performance degradation.

server/schemas/item.graphql
mimoham24: @rot1024 during the discussion with Baba-chan he said that it might be needed in future
however I'll remove it for now

server/internal/infrastructure/mongo/item.go
```diff
@@ -20,7 +21,16 @@ type itemRepo struct {
...
 
 func NewItem(client *mongox.Client) repo.Item {
-	return &itemRepo{client: mongogit.NewCollection(client.WithCollection("item"))}
+	r := &itemRepo{client: mongogit.NewCollection(client.WithCollection("item"))}
+	r.init()
+	return r
+}
+
+func (r *itemRepo) init() {
+	err := r.client.CreateIndexes(context.Background(), []string{"schema", "fields.value"}, []string{"id", "fields.schemafield"})
```
rot1024, line 30:
```suggestion
	err := r.client.CreateIndexes(context.Background(), []string{"schema", "fields.schemafield"}, []string{"id"})
```

server/schemas/item.graphql
```diff
@@ -1,6 +1,9 @@
 type Item implements Node {
   id: ID!
   schemaId: ID!
+  projectId: ID!
+  project: Project!
+  schema: Schema!
   fields: [ItemField!]
```
rot1024, line 7:
```suggestion
  fields: [ItemField!]!
```

rot1024 (changes requested):

rot1024 (approved):

PR #282, opened 2022-10-13
feat(web): add comments hooks

issmail-basel:


issmail-basel:
Already merged and added in other PR

PR #295, opened 2022-10-18
fix(server): jwt echo middleware

nourbalaha:
# Overview
In this PR I fixed JWT echo middleware to be sent with rest api requests.

## What I have done
I changed 
`appx.AuthMiddleware(cfg.Config.JWTProviders(), adapter.ContextAuthInfo, false)`
into
`appx.AuthMiddleware(cfg.Config.JWTProviders(), adapter.ContextAuthInfo, true)`
and wrapped the middleware inside `jwtEchoMiddleware` function.

### DIFF
--- a/server/internal/app/auth_client.go
+++ b/server/internal/app/auth_client.go
@@ -10,6 +10,7 @@ import (
 	"github.com/reearth/reearth-cms/server/internal/usecase/interfaces"
 	"github.com/reearth/reearth-cms/server/pkg/id"
 	"github.com/reearth/reearth-cms/server/pkg/user"
+	"github.com/reearth/reearthx/appx"
 	"github.com/reearth/reearthx/usecasex"
 	"github.com/samber/lo"
 )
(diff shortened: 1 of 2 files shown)


server/internal/app/auth_client.go
```diff
@@ -53,6 +54,14 @@ func authMiddleware(cfg *ServerConfig) echo.MiddlewareFunc {
 	}
 }
 
+func jwtEchoMiddleware(cfg *ServerConfig) echo.MiddlewareFunc {
+	mw, err := appx.AuthMiddleware(cfg.Config.JWTProviders(), adapter.ContextAuthInfo, true)
+	if err != nil {
+		panic(err)
+	}
```
rot1024, line 61:
```suggestion
	mw := lo.Must(appx.AuthMiddleware(cfg.Config.JWTProviders(), adapter.ContextAuthInfo, true))
```

I used `lo.Must` to summarize it in one line, was it hard to read so that you created a new function?

rot1024 (approved):

server/internal/app/auth_client.go
nourbalaha: I thought that it looks cleaner this way.
Anyway I will use `lo.Must` to summarize it as you mentioned above. 
Thank you!

PR #303, opened 2022-10-20
fix(web): members section fix

issmail-basel:
# Fixes:
- [ ]  UI shouldn’t collapse when user name is long 
- [ ]  Even after a member has been added to the workspace, and next time to try to add another user, modal still shows the user who has been already added 
- [ ]  If you zoom the development tool to the left, the style is broken. (part of action, change role) → css  broken 
- [ ]  If we try to invite a member who already members, UI doesn’t show the error “already exist” 
- [x]  When User is a “reader”, all buttons in the members page will be disable.
- [ ]  The order of members should be consistent ( now it’s random )
- [ ]  “OK” should be “Invite”
- [ ]  Only when user is found, “Invite” button should be active
- [x]  User shouldn’t be able to search for user self


web/src/gql/provider.tsx
```diff
@@ -37,6 +38,7 @@ const Provider: React.FC<Props> = ({ children }) => {
     if (!networkError && !graphQLErrors) return;
     const error = networkError?.message ?? graphQLErrors?.map(e => e.message).join(", ");
     if (error) {
+      Notification.error({ message: error });
```
KaWaite, line 41: Since you are using Notification.error in places of errors, I think you can remove the setError below this and remove that completely since I dont think we are using that then.

web/src/components/organisms/Settings/Members/index.tsx
```diff
@@ -126,7 +133,11 @@ const Members: React.FC = () => {
         <ActionHeader>
           <Search placeholder={t("search for a member")} allowClear style={{ width: 264 }} />
         </ActionHeader>
-        <Table dataSource={dataSource} columns={columns} style={{ padding: "24px" }} />
+        <Table
+          dataSource={dataSource}
+          columns={columns}
+          style={{ padding: "24px", overflow: "auto" }}
```
KaWaite, line 139: For the table we just want scroll on the horizontal pane, right? So please use overflowX

web/src/components/organisms/Settings/Members/hooks.ts
```diff
@@ -73,7 +96,8 @@ export default ({ workspaceId }: Props) => {
             }
           : undefined,
       )
-      .filter((user): user is Member => !!user);
+      .filter((user): user is Member => !!user)
+      .sort((a, b) => stringSortCallback(a.userId, b.userId));
```
KaWaite, line 100: a and b don't have any meaning that I can see. Can you rename to user1 and user2?

web/src/components/organisms/Settings/Members/hooks.ts
```diff
@@ -37,6 +39,23 @@ export default ({ workspaceId }: Props) => {
...
   const workspaces = data?.me?.workspaces as Workspace[];
 
+  const checkOwner = useCallback(() => {
+    const members = currentWorkspace?.members;
+    if (members) {
+      for (let i = 0; i < members.length; i++) {
+        if (members[i].userId === me?.id && members[i].role === "OWNER") {
+          return true;
+        }
+      }
```
KaWaite, line 49: Please do something like this (please test this and fix if I have any errors in my suggestion):
```
const isOwner = useMemo(()=> currentWorkspace?.members.find(m => m.userId === me?.id && m.role === "OWNER", [currentWorkspace?.members, me?.id])
```

web/src/components/organisms/Settings/Members/hooks.ts
```diff
@@ -37,6 +39,23 @@ export default ({ workspaceId }: Props) => {
...
+    }
+    return false;
+  }, [currentWorkspace?.members, me?.id]);
+
+  useEffect(() => {
+    const o = checkOwner();
+    setOwner(o);
```
KaWaite, line 56: setOwner(isOwner());

KaWaite (commented):
Reviewed

web/src/gql/provider.tsx
issmail-basel: https://github.com/reearth/reearth-cms/blob/d5369fdbd6bba4c3caf029b5daf5e77166d37ac5/web/src/gql/provider.tsx#L38-L42
Removed it, with all it's occurrences in the `state` file

web/src/components/organisms/Settings/Members/index.tsx
issmail-basel: Fixed.

web/src/components/organisms/Settings/Members/hooks.ts
issmail-basel: Fixed.

web/src/components/organisms/Settings/Members/hooks.ts
issmail-basel: Fixed, and tested.

web/src/components/organisms/Settings/Members/hooks.ts
issmail-basel: Changed to `setOwner(isOwner)`

KaWaite (approved):

PR #305, opened 2022-10-20
fix(web): project section fix 

issmail-basel:
# Fixes:
- [ ]  FE: After user edit the project settings, FE should update the cache and show updated info 
- [ ]  FE: When Reader tried to change the project settings, text-area becomes blank


web/src/components/organisms/Project/settings/hooks.ts
```diff
@@ -8,14 +8,15 @@ import {
...
-import { useWorkspace } from "@reearth-cms/state";
+import { useWorkspace, useProject } from "@reearth-cms/state";
 
 type Params = {
   projectId?: string;
 };
 
 export default ({ projectId }: Params) => {
   const navigate = useNavigate();
+  const [_, setCurrentProject] = useProject();
```
KaWaite, line 19:
```suggestion
  const [, setCurrentProject] = useProject();
```

web/src/components/organisms/Project/settings/hooks.ts
```diff
@@ -62,9 +63,11 @@ export default ({ projectId }: Params) => {
         Notification.error({ message: t("Failed to update project.") });
         return;
       }
+      const updatedProjectData = project.data.updateProject.project;
+      setCurrentProject({ ...updatedProjectData });
```
KaWaite, line 67: I was trying to handle setting currentProject in the header, and only the header. 
Can you add a refetch to the updateProject mutation here and then take a look there to see if you can add/edit a dependency in the useEffect in the header that would update the currentProject after the cache is updated?

web/src/components/organisms/Project/settings/index.tsx
```diff
@@ -44,7 +44,6 @@ const ProjectSettings: React.FC = () => {
           name: values.name,
           description: values.description,
         });
-        form.resetFields();
```
KaWaite, line 47: When you remove this, doesn't this mean after submitting the form will NEVER reset?

KaWaite (commented):
Reviewed

web/src/components/organisms/Project/settings/hooks.ts
issmail-basel: @KaWaite ,  nice, I got used to `underscore` for unused variables, but this did not come to my mind.

web/src/components/organisms/Project/settings/hooks.ts
issmail-basel: @KaWaite , these lines:
https://github.com/reearth/reearth-cms/blob/273a9e513780a133e01099a01050682b3b3a3479/web/src/components/organisms/CMSWrapper/hooks.ts#L94
will prevent refetch (because the `projectId` is equal to `currentProject?.id` after the update) 
and:
https://github.com/reearth/reearth-cms/blob/273a9e513780a133e01099a01050682b3b3a3479/web/src/components/organisms/CMSWrapper/hooks.ts#L99

Will prevent updating the state for same reason, any suggestions, and what the reason for these lines?

web/src/components/organisms/Project/settings/index.tsx
issmail-basel: @KaWaite , yes and this is intentional because it is an update form, it should keep the same of info.

web/src/components/organisms/Project/settings/hooks.ts
KaWaite: Can you actually remove those checks that are checking if projectId equals/doesn't equal currentProject, then do as I suggested above with the refetch?
I added those checks so that there weren't unneeded refetches or updating the currentProject, but I realize they are maybe unnecessary/not good.

issmail-basel: I tried to run the code without it, doesn't seem to affect.
I have updated the code for those parts.

KaWaite: Is your linter working correctly? Since you removed the currentProject?.id check from the useEffect, you should be getting warnings to remove the dependency since its not referenced any more in the useEffect

issmail-basel: I'll check and get back

KaWaite (approved):

PR #317, opened 2022-10-22
feat(web): add workspace settings and user account

issmail-basel:


web/src/App.tsx
```diff
@@ -17,85 +15,77 @@ import Members from "@reearth-cms/components/organisms/Settings/Members";
 import MyIntegrationDetails from "@reearth-cms/components/organisms/Settings/MyIntegrationDetails";
 import MyIntegrations from "@reearth-cms/components/organisms/Settings/MyIntegrations";
 import Workspace from "@reearth-cms/components/organisms/Workspace";
+import WorkspaceSettings from "@reearth-cms/components/organisms/Workspace/Settings";
 import CMSPageWrapper from "@reearth-cms/components/pages/CMSPage";
 import RootPage from "@reearth-cms/components/pages/RootPage";
 import { Provider as GqlProvider } from "@reearth-cms/gql";
 import { Provider as I18nProvider } from "@reearth-cms/i18n";
 
+import AccountSettings from "./components/organisms/Account";
```
KaWaite, line 24: Path should be from @reearht-cms

web/src/components/organisms/Workspace/Settings/index.tsx
```diff
@@ -0,0 +1,17 @@
+import WorkspaceSettingsMolecule from "@reearth-cms/components/molecules/WorkspaceSettings";
+
+import useHooks from "./hooks";
+
+const WorkspaceSettings: React.FC = () => {
+  const { handleWorkspaceUpdate, handleWorkspaceDelete, workspaceName } = useHooks();
```
KaWaite, line 6: I know its small, but please try to keep imports, props, etc in the same order as their type or passed order. In this case, workspaceName is returned first in the hooks file, so should be first here.
Additionally, if you can keep functions AFTER variables, that would be perfect (just like here will be once workspaceName is moved)

web/src/components/organisms/Workspace/Settings/hooks.ts
```diff
@@ -0,0 +1,61 @@
...
+  const [currentWorkspace, setCurrentWorkspace] = useWorkspace();
+  const t = useT();
+
+  const workspaceId = currentWorkspace?.id;
+  const workspaceName = currentWorkspace?.name;
+
+  const [updateWorkspaceMutation] = useUpdateWorkspaceMutation();
+  const [deleteWorkspaceMutation] = useDeleteWorkspaceMutation({
+    refetchQueries: ["GetMe"],
+  });
```
KaWaite, line 23: Does the updateWorkspaceMutation not also need to refetch?

web/src/components/organisms/Workspace/Settings/hooks.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const handleWorkspaceUpdate = useCallback(
+    async (name?: string) => {
+      if (!workspaceId || !name) return;
+      const workspace = await updateWorkspaceMutation({
```
KaWaite, line 28: `const resp` or `res` is more accurate and anyways common.

web/src/components/organisms/Workspace/Settings/hooks.ts
```diff
@@ -0,0 +1,61 @@
...
+          workspaceId,
+          name,
+        },
+      });
+      if (workspace.errors || !workspace.data?.updateWorkspace) {
+        Notification.error({ message: t("Failed to update workspace.") });
+        return;
+      }
+      setCurrentWorkspace({ ...workspace.data.updateWorkspace.workspace });
+      Notification.success({ message: t("Successfully updated workspace!") });
```
KaWaite, line 39: I think I have let through this quite often, so I apologize for suddenly asking for a change, but I feel like if/else is more clear and understandable then using the return. Can you update where you wrote this in this PR and then while doing other tasks if you come across this code can you slowly update?

web/src/components/organisms/Account/hooks.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (!me) return;
+    const user = await deleteMeMutation({ variables: { userId: me.id } });
+    if (user.errors) {
+      Notification.error({ message: t("Failed to delete user.") });
+      return;
+    } else {
+      Notification.success({ message: t("Successfully deleted user!") });
+      logout();
+    }
+  }, [me, deleteMeMutation, logout, t]);
```
KaWaite, line 74: Have you tried deleting yourself to make sure this is working?

web/src/components/organisms/Account/hooks.ts
```diff
@@ -0,0 +1,82 @@
...
+      }
+      Notification.success({ message: t("Successfully updated user!") });
+    },
+    [updateMeMutation, t],
+  );
+
+  const handleLanguageUpdate = useCallback(
+    async (lang?: string) => {
+      if (!lang) return;
+      const language = await updateMeMutation({ variables: { lang } });
```
KaWaite, line 52: Please update anywhere else its like this. Should be `const res` since what you are getting is a response, not the language

web/src/components/organisms/Account/hooks.ts
```diff
@@ -0,0 +1,82 @@
...
+      if (language.errors) {
+        Notification.error({ message: t("Failed to update language.") });
+        return;
+      } else {
+        await client.resetStore();
```
KaWaite, line 57: Is there a reason you need this here? I feel like on a language update you shouldn't need this

web/src/components/molecules/WorkspaceSettings/Form.tsx
```diff
@@ -0,0 +1,48 @@
...
+      style={{ maxWidth: 400 }}
+      form={form}
+      initialValues={{ name: workspaceName }}
+      layout="vertical"
+      autoComplete="off">
+      <Form.Item
+        name="name"
+        label={t("Workspace Name")}
+        extra={t(
+          "This is your team's visible name within Re:Earth and Re:Earth CMS. For example, the name of your company or department.",
```
KaWaite, line 37: This is the name that will be visible within Re:Earth and Re:Earth CMS. This could be your company's name, department's name, the theme of your projects, etc.

web/src/components/molecules/WorkspaceSettings/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  workspaceName,
+  onWorkspaceUpdate,
+  onWorkspaceDelete,
+}) => {
+  const t = useT();
+
+  return (
+    <InnerContent title="Workspace setting">
+      <ContentSection title={t("General")}>
+        <WorkspaceForm onWorkspaceUpdate={onWorkspaceUpdate} workspaceName={workspaceName} />
```
KaWaite, line 23: Please start to pay attention to prop order and keep it the same as the Props type of the component you are passing to.
Like I mentioned in another PR's comment, try to keep variables first, followed by functions.

web/src/components/molecules/WorkspaceSettings/Form.tsx
```diff
@@ -0,0 +1,48 @@
...
+const WorkspaceForm: React.FC<Props> = ({ workspaceName, onWorkspaceUpdate }) => {
+  const [form] = Form.useForm();
+  const t = useT();
+
+  const handleSubmit = useCallback(async () => {
+    try {
+      const values = await form.validateFields();
+      await onWorkspaceUpdate?.(values.name);
+    } catch (info) {
+      console.log("Validate Failed:", info);
```
KaWaite, line 22: Can you handle this as a notification? Or if its not necessary because antd's form will handle the feedback, remove the try/catch

web/src/components/atoms/ContentSection/index.tsx
```diff
@@ -0,0 +1,50 @@
+import styled from "@emotion/styled";
+import { ReactNode } from "react";
+
+export type Props = {
+  title: string;
+  danger?: boolean;
+  extra?: ReactNode;
```
KaWaite, line 7: What is the extra for?

web/src/components/atoms/ContentSection/index.tsx
```diff
@@ -0,0 +1,50 @@
+import styled from "@emotion/styled";
+import { ReactNode } from "react";
+
+export type Props = {
+  title: string;
+  danger?: boolean;
```
KaWaite, line 6: I don't think the danger is necessary, and makes this atom less reuseable. Please change this to `style` and then pass the border in where necessary.

web/src/components/atoms/ContentSection/index.tsx
```diff
@@ -0,0 +1,50 @@
+import styled from "@emotion/styled";
+import { ReactNode } from "react";
+
+export type Props = {
```
KaWaite, line 4: You are missing the `className?: string` prop. Without it, this can't be used with styling libraries like Styled Components

web/src/components/atoms/ContentSection/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  danger?: boolean;
+  extra?: ReactNode;
+  children?: ReactNode;
+};
+
+const ContentSection: React.FC<Props> = ({ title, extra, children, danger }) => {
+  return (
+    <Wrapper danger={danger}>
+      <Header>
+        <Subtitle>{title}</Subtitle>
```
KaWaite, line 15: In the context of this component, this is a Title, not Subtitle

web/src/components/atoms/ContentSection/index.tsx
```diff
@@ -0,0 +1,50 @@
+import styled from "@emotion/styled";
```
KaWaite, line 1: Please move ContentSection in to the InnerContents directory as it is a child component that is suppose to be used together with InnerContents

web/src/components/molecules/AccountSettings/DangerZone.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <Text>
+        {t(
+          "Permanently remove your Personal Account and all of its contents from the Re:Earth CMS. This action is not reversible, so please continue with caution.",
+        )}
+      </Text>
+
+      <Button onClick={handleAccountDeleteConfirmation} type="primary" danger>
+        {t("Delete Personal Account")}
+      </Button>
+    </>
```
KaWaite, line 39: In this component can you use the ContentSection? Without the ContentSection, this component has no structure and becomes a little useless. If we decide to move the Dangerzone, for example, if we don't know this needs to be wrapped in the ContentSection, it will be messy.

web/src/components/molecules/AccountSettings/ServiceForm.tsx
```diff
@@ -0,0 +1,65 @@
...
+    [t],
+  );
+
+  const handleSubmit = useCallback(async () => {
+    try {
+      const values = await form.validateFields();
+      await onLanguageUpdate?.(values.lang);
+    } catch (info) {
+      console.log("Validate Failed:", info);
+    }
```
KaWaite, line 36: Same as elsewhere I commented, remove the try/catch if not needed

web/src/components/molecules/AccountSettings/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+const AccountSettings: React.FC<Props> = ({
+  user,
+  onUserDelete,
+  onLanguageUpdate,
+  onUserUpdate,
+}) => {
+  const t = useT();
+
+  return (
+    <InnerContent title="Workspace setting">
```
KaWaite, line 25: Workspace Settings

web/src/components/molecules/AccountSettings/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <ContentSection title={t("General")}>
+        <AccountGeneralForm user={user} onUserUpdate={onUserUpdate} />
+      </ContentSection>
+      <ContentSection title={t("Service")}>
+        <AccountServiceForm user={user} onLanguageUpdate={onLanguageUpdate} />
+      </ContentSection>
+      <ContentSection title={t("Dangerous Zone")} danger>
```
KaWaite, line 32: Danger Zone

web/src/components/molecules/WorkspaceSettings/DangerZone.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <br />
+        {t(
+          "So when you delete or create a workspace in service A and service B, they will all be synchronized with each other. Make sure you understand what you're doing.",
+        )}
+      </Text>
+
+      <Button onClick={handleWorkspaceDeleteConfirmation} type="primary" danger>
+        {t("Remove Workspace")}
+      </Button>
+    </>
```
KaWaite, line 43: This is the old design. Please double check the figma file and update accordingly!

issmail-basel: @KaWaite , Red have updated the Figma to have this design for all settings pages.

web/src/App.tsx
issmail-basel: Changed.

web/src/components/organisms/Workspace/Settings/index.tsx
issmail-basel: @KaWaite , I'll keep this in mind.

web/src/components/organisms/Workspace/Settings/hooks.ts
issmail-basel: @KaWaite , no need, I'm updating the state of workspace in case of success.
In case of delete I need to remove the workspace from the list, that's why I refetch.

web/src/components/organisms/Workspace/Settings/hooks.ts
issmail-basel: @KaWaite , In this case in particular I disagree, I feel the `return` style of coding when something is wrong is better than `if/else`
Should we take advice and ask Inouse-san?
If not let me know, and I'll change it.

web/src/components/organisms/Account/hooks.ts
issmail-basel: @KaWaite , yes it deletes the user from `mongoDB`.
But on the local system, it will create the user again after logging in (because of Baba-chan change in our local environment)

web/src/components/organisms/Account/hooks.ts
issmail-basel: Updated.

web/src/components/organisms/Account/hooks.ts
issmail-basel: @KaWaite , no need, removed.

web/src/components/molecules/WorkspaceSettings/Form.tsx
issmail-basel: @KaWaite , edited.

web/src/components/molecules/WorkspaceSettings/index.tsx
issmail-basel: @KaWaite , changed, and I'll keep this in mind for future.

web/src/components/molecules/WorkspaceSettings/Form.tsx
issmail-basel: @KaWaite , no need for `try/catch` after I've added notification to hooks/error handler.

web/src/components/atoms/ContentSection/index.tsx
issmail-basel: @KaWaite , inspired by `extra` in antd design, used if you need actions/content on the other side of the title.

web/src/components/atoms/ContentSection/index.tsx
issmail-basel: @KaWaite , it only uses two states, the normal and red bordered, so should I made it as `style` and pass, or is it an over burn for this case?

web/src/components/atoms/ContentSection/index.tsx
issmail-basel: @KaWaite , added.
Did not know that I encountered a place where I couldn't use `styled components` but It never crossed my mind that this is the reason why.

web/src/components/atoms/ContentSection/index.tsx
issmail-basel: @KaWaite , you are right.
Fixed.

web/src/components/atoms/ContentSection/index.tsx
issmail-basel: @KaWaite , done.

web/src/components/molecules/AccountSettings/index.tsx
issmail-basel: @KaWaite , updated.

web/src/components/molecules/AccountSettings/ServiceForm.tsx
issmail-basel: @KaWaite , updated, same reason as I stated above as well.

web/src/components/molecules/AccountSettings/ServiceForm.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <Form
+      style={{ maxWidth: 400 }}
+      form={form}
+      initialValues={user}
+      layout="vertical"
+      autoComplete="off">
+      <Form.Item
+        name="lang"
+        label={t("Service Language")}
+        extra={t("This will change the UI langauge")}>
```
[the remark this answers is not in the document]
issmail-basel: @KaWaite , updated.

web/src/components/molecules/AccountSettings/GeneralForm.tsx
```diff
@@ -0,0 +1,53 @@
...
+    <Form
+      style={{ maxWidth: 400 }}
+      form={form}
+      initialValues={user}
+      layout="vertical"
+      autoComplete="on">
+      <Form.Item
+        name="name"
+        label={t("Account Name")}
+        extra={t("This is your id in Re:Earth and Re:Earth CMS")}>
```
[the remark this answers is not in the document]
issmail-basel: @KaWaite , updated.

web/src/components/molecules/AccountSettings/DangerZone.tsx
issmail-basel: @KaWaite , you are right, updated and noted.

web/src/components/organisms/Account/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+      {me && (
+        <AccountSettingsMolecule
+          user={me}
+          onUserUpdate={handleUserUpdate}
+          onLanguageUpdate={handleLanguageUpdate}
+          onUserDelete={handleUserDelete}
+        />
+      )}
+    </>
+  );
```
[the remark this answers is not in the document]
KaWaite: This is being passed to the CMSWrapper isn't it? So returning undefined would be fine. But you are showing account settings if there is a me, so wouldn't it be better to show the Loading component if loading/!me and then show the AZccountSettingsMolecule when loading is done and/or there is a me?

web/src/components/organisms/Workspace/Settings/hooks.ts
KaWaite: That is fine for displaying the workspace settings page, but what about the workspace list in the header? Does the name there update if you don't refetch?

web/src/components/organisms/Workspace/Settings/hooks.ts
KaWaite: But workspace isn't what is represented here. This is a response object with various fields.

web/src/components/organisms/Workspace/Settings/hooks.ts
KaWaite: @issmail-basel Honestly both get the job done. Just in this case I think we usually do if/else, and you are actually doing if/else in the handleWorkspaceDelete function right below this function. 
Please change here and we can discuss more later in the week if you want to.

web/src/components/atoms/ContentSection/index.tsx
KaWaite: If we don't need it in the design, please remove it. We can add things like that when the need arises

web/src/components/atoms/ContentSection/index.tsx
KaWaite: Hmmmmm I'll leave this up to you. After you make a decision you can resolve this conversation.

web/src/components/atoms/ContentSection/index.tsx
KaWaite: Yep. I think recently I fixed an issue with the Icon atom because we were trying to do styled components with it and nothing was happening because it had no className!
(once you read this, you can resolve this convo)

web/src/components/molecules/AccountSettings/DangerZone.tsx
```diff
@@ -0,0 +1,57 @@
...
+      },
+    });
+  }, [confirm, onUserDelete, t]);
+
+  return (
+    <ContentSection title={t("Danger Zone")} danger>
+      <Title>{t("Delete Personal Account")}</Title>
+      <Text>
+        {t(
+          "Permanently remove your Personal Account and all of its contents from the Re:Earth CMS. This action is not reversible, so please continue with caution.",
```
KaWaite, line 33: Permanently removes your personal account and all of its contents from Re:Earth CMS. This action is not reversible, so please continue with caution.

web/src/components/organisms/Workspace/Settings/hooks.ts
issmail-basel: @KaWaite , yes it is updated.

web/src/components/organisms/Workspace/Settings/hooks.ts
issmail-basel: @KaWaite , sure, I changed it.

web/src/components/organisms/Account/hooks.ts
issmail-basel: @KaWaite , fixed.

web/src/components/atoms/ContentSection/index.tsx
issmail-basel: @KaWaite , it is used by the `Project overview` page, that's why it is added.

web/src/components/atoms/ContentSection/index.tsx
issmail-basel: @KaWaite , I think for now it is enough, no other border color or style changes will occur, it is better to keep the options limited and clear for now.

web/src/components/organisms/Workspace/Settings/hooks.ts
```diff
@@ -0,0 +1,61 @@
...
+      const res = await updateWorkspaceMutation({
+        variables: {
+          workspaceId,
+          name,
+        },
+      });
+      if (res.errors || !res.data?.updateWorkspace) {
+        Notification.error({ message: t("Failed to update workspace.") });
+      } else {
+        setCurrentWorkspace({ ...res.data.updateWorkspace.workspace });
```
KaWaite, line 37: Can't you just do `setCurrentWorkspace(res.data.update.workspace.workspace)`?

web/src/components/molecules/AccountSettings/index.tsx
KaWaite: Settings should be capitalized too

web/src/components/molecules/WorkspaceSettings/DangerZone.tsx
KaWaite: @issmail-basel Please check again. I talked with him the other day because the language here is confusing to figure out what he wanted to say, and it turned out there were two versions. The one you have here was an older version that he removed, so please check the file again

web/src/components/atoms/ContentSection/index.tsx
KaWaite: So there are a couple issues with the current implementation. First, the name just doesn't really convey the meaning of what it is. Right now I can't think of a better name, so if you can think of something more meaningful, please suggest it. 
Second, the Title inside ContentSection has no padding or margin, and the Header doesn't wrap, which means the extra prop's component isn't really going to work nicely if it is long or the browser width becomes a little narrow. 

If you can solve the above issues your approach is fine. Another option would be to make Title `string | ReactNode`, then you could make the Header fully customizable.

web/src/components/molecules/AccountSettings/DangerZone.tsx
KaWaite: Still not addressed.

web/src/components/molecules/WorkspaceSettings/DangerZone.tsx
```diff
@@ -0,0 +1,59 @@
...
+      },
+    });
+  }, [confirm, onWorkspaceDelete, t]);
+
+  return (
+    <ContentSection title={t("Danger Zone")} danger>
+      <Title>{t("Remove Workspace")}</Title>
+      <Text>
+        {t(
+          "Permanently remove your Workspace and all of its contents from the Re:Earth CMS. This action is not reversible – please continue with caution.",
```
KaWaite, line 33: Permanently removes the Workspace and all of its contents from Re:Earth CMS. This action is not reversible, so please continue with caution.

web/src/components/organisms/Workspace/Settings/hooks.ts
KaWaite: Still not addressed

web/src/components/atoms/ProTable/index.tsx
```diff
@@ -1,9 +1,34 @@
...
 
-export default ProTable;
+import { useLang } from "@reearth-cms/i18n";
+
+export type Props = ProTableProps<Record<string, any> | any, ParamsType, "text">;
+
+const Table: React.FC<Props> = props => {
+  const [antdLang, setAntdLang] = useState<Locale>(enUSIntl);
+  const lang = useLang();
+  console.log(lang);
```
KaWaite, line 18: Don't need

web/src/components/molecules/AccountSettings/DangerZone.tsx
issmail-basel: @KaWaite , updated.

web/src/components/organisms/Workspace/Settings/hooks.ts
issmail-basel: @KaWaite , yes it can be done, updated.

web/src/components/atoms/ProTable/index.tsx
issmail-basel: @KaWaite , fixed.

web/src/components/molecules/WorkspaceSettings/DangerZone.tsx
issmail-basel: @KaWaite , updated.

web/src/components/atoms/ProTable/index.tsx
issmail-basel: https://github.com/reearth/reearth-cms/blob/8b30aeb777912d1b1a63a86183ee255169c9329e/web/src/components/atoms/ProTable/index.tsx#L13-L21

KaWaite (approved):

PR #320, opened 2022-10-24
feat(server): add unknown preview type

nourbalaha:
# Overview
In this PR I added "UNKNOWN" to asset's preview type


server/pkg/asset/preview_type.go
```diff
@@ -13,6 +13,7 @@ const (
 	PreviewTypeGEO     PreviewType = "GEO"
 	PreviewTypeGEO3D   PreviewType = "GEO3D"
 	PreviewTypeMODEL3D PreviewType = "MODEL3D"
+	PreviewTypeUNKNOWN PreviewType = "UNKNOWN"
```
rot1024, line 16:
```suggestion
	PreviewTypeUNKNOWN PreviewType = ""
```

Zero value is suitable, I think.

server/pkg/asset/preview_type.go
```diff
@@ -25,6 +26,8 @@ func PreviewTypeFrom(p string) (PreviewType, bool) {
 		return PreviewTypeGEO3D, true
 	case PreviewTypeMODEL3D:
 		return PreviewTypeMODEL3D, true
+	case PreviewTypeUNKNOWN:
+		return PreviewTypeUNKNOWN, true
 	default:
 		return PreviewType(""), false
 	}
```
rot1024, line 33: Use `strings.ToLower` for `p` in this method

server/pkg/asset/preview_type.go
```diff
@@ -44,6 +47,8 @@ func PreviewTypeFromRef(p *string) *PreviewType {
 		p2 = PreviewTypeGEO3D
 	case PreviewTypeMODEL3D:
 		p2 = PreviewTypeMODEL3D
+	case PreviewTypeUNKNOWN:
+		p2 = PreviewTypeUNKNOWN
 	default:
 		return nil
```
rot1024, line 53: Implementation of this func is duplicated with `PreviewTypeFrom`.  Use `PreviewTypeFrom(*p)` instead.

server/pkg/asset/preview_type.go
```diff
@@ -13,6 +13,7 @@ const (
 	PreviewTypeGEO     PreviewType = "GEO"
 	PreviewTypeGEO3D   PreviewType = "GEO3D"
 	PreviewTypeMODEL3D PreviewType = "MODEL3D"
```
rot1024, line 15: Please lowercase all existing constant strings. The use of uppercase is for GraphQL convenience and is not common for constants used by Re:Earth. There is no impact because the conversion logic has already been properly implemented.

rot1024 (changes requested):

server/pkg/asset/preview_type.go
nourbalaha: I am getting a lot of errors because of this change,
now there is come confusion between `(*PreviewType)(nil)` and `(*PreviewType)("")`
also the `PreviewTypeUnknown` is "unknown" and not "" in gql and the adapter level
so is it okay not to do this change?

server/pkg/asset/preview_type.go
nourbalaha: I have already deleted this method since it is not used

server/pkg/asset/preview_type.go
nourbalaha: Done!

server/pkg/asset/preview_type.go
nourbalaha: Done!

server/pkg/asset/preview_type.go
```diff
@@ -9,41 +9,29 @@ import (
...
-		return PreviewTypeGEO3D, true
-	case PreviewTypeMODEL3D:
-		return PreviewTypeMODEL3D, true
-	default:
-		return PreviewType(""), false
-	}
-}
-
-func PreviewTypeFromRef(p *string) *PreviewType {
+func PreviewTypeFrom(p *string) *PreviewType {
```
rot1024, line 19: I think it is good that `PreviewTypeFrom` and `PreviewTypeFromRef` are available since `lo.ToPtr` can be reduced.

server/pkg/asset/preview_type.go
```diff
@@ -53,9 +41,9 @@ func PreviewTypeFromRef(p *string) *PreviewType {
 
 func PreviewTypeFromContentType(c string) *PreviewType {
```
rot1024, line 42: I don't think it's unreasonable that everything should be geo except images; if it's not determined to be geo, it should be unknown. identification by extension rather than content-type might be better.

rot1024 (changes requested):

server/pkg/asset/preview_type.go
rot1024: https://github.com/reearth/reearth-cms/blob/main/server/internal/adapter/gql/resolver_mutation_asset.go#L42

This line occurs a bug that GraphQL representation is uppercase, but domain model's representation is lowercase. So we need a function to convert GraphQL's asset preview type to domain model's asset preview type.

server/pkg/asset/preview_type.go
nourbalaha: `PreviewTypeFrom` is not used anywhere that is why I deleted it, should I rename PreviewTypeFrom to PreviewTypeFromRef?

server/pkg/asset/preview_type.go
nourbalaha: Agree!

server/pkg/asset/preview_type.go
rot1024: I saw `PreviewTypeFrom(lo.ToPtr(hogehoge))` at https://github.com/reearth/reearth-cms/pull/320/files#diff-5973a8c58a7fed17669414bc79a16b0bd282de7f6cf6050d7333b88f73230c67R372, for example.

server/pkg/asset/preview_type.go
nourbalaha: Done!

server/pkg/asset/preview_type.go
nourbalaha: I thought that this is happening in the adapter layer by this function:
https://github.com/reearth/reearth-cms/blob/feat/add-unknown-preview-type/server/internal/adapter/gql/gqlmodel/convert_asset.go#L32

server/pkg/asset/preview_type.go
nourbalaha: Done!

server/pkg/asset/preview_type.go
rot1024: It is domain model -> graphql model, but I say about graphql mode -> domain model.

server/pkg/asset/preview_type.go
rot1024: Do not forget fixing it. Implementation duplicated.

server/pkg/asset/preview_type.go
nourbalaha: Done!

server/pkg/asset/preview_type.go
nourbalaha: I am not sure if I get what you mean, can you please explain when you have time. Thank you!

nourbalaha: Done!

rot1024 (approved):

PR #326, opened 2022-10-26
feat(web): make content table re-sizable columns

issmail-basel:


web/src/components/organisms/Project/Content/ContentList/hooks.ts
```diff
@@ -32,6 +32,9 @@ export default () => {
       title: field.title,
       dataIndex: ["fields", field.id],
       key: field.id,
+      width: 128,
+      minWidth: 128,
```
KaWaite, line 36: Unless there is a weird issue with AntD, `width` should be enough, no?

web/src/index.css
```diff
@@ -8,4 +8,19 @@ body {
...
+    bottom: 0;
+    z-index: 1;
+    width: 10px;
+    height: 100%;
+    cursor: col-resize;
+}
+
+.react-resizable {
+    position: relative;
+    background-clip: padding-box;
```
KaWaite, line 25: Please just pass these in to the component that uses them since they aren't truly global and we always pass the antd class modifications in to the component directly

web/src/components/molecules/Common/ResizableProTable/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+  options,
+  toolbar,
+}) => {
+  const [columns, setColumns] = useState<ProColumns<any, "text">[]>([]);
+
+  useEffect(() => {
+    if (Array.isArray(proColumns)) {
+      setColumns(proColumns);
+    }
+  }, [proColumns, setColumns]);
```
KaWaite, line 34: Is there a specific reason you are checking if it is an array and not just checking if it exists (if(proColumns))?

web/src/components/molecules/Common/ResizableProTable/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+        width: size.width,
+      };
+      setColumns(newColumns);
+    };
+
+  const mergeColumns: ProColumns<any, "text">[] = columns?.map((col, index) => ({
+    ...col,
+    onHeaderCell: column => ({
+      minWidth: (column as ProColumns<any, "text"> & { minWidth: number }).minWidth,
+      width: (column as ProColumns<any, "text">).width,
```
KaWaite, line 51: Do you need minWidth if you are setting width? Typically if you set width it'll be that size always

web/src/components/molecules/Common/ResizableProTable/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+        }}
+        rowKey="id"
+        search={false}
+        loading={loading}
+        toolbar={toolbar}
+        options={options}
+        tableStyle={{ overflowX: "scroll" }}
+      />
+    </Wrapper>
+  ) : null;
```
KaWaite, line 74: Do you want to return null? I thought we would want to still show the table even if no values and let the table display the antd `No data` message like the other tables, no?

web/src/components/molecules/Common/ResizableProTable/resizble.tsx
```diff
@@ -0,0 +1,36 @@
+import { Resizable } from "react-resizable";
+import type { ResizeCallbackData } from "react-resizable";
+
+export type { ResizeCallbackData } from "react-resizable";
+
+export const ResizableTitle = (
+  props: React.HTMLAttributes<any> & {
+    onResize: (e: React.SyntheticEvent<Element>, data: ResizeCallbackData) => void;
+    width: number;
+    minWidth: number;
```
KaWaite, line 10: Both necessary?

web/src/components/molecules/Common/ResizableProTable/resizble.tsx
```diff
@@ -0,0 +1,36 @@
+import { Resizable } from "react-resizable";
+import type { ResizeCallbackData } from "react-resizable";
+
+export type { ResizeCallbackData } from "react-resizable";
+
+export const ResizableTitle = (
```
KaWaite, line 6: This file's name is wrong.
resizble.tsx -> resizable.tsx

web/src/components/organisms/Project/Content/ContentList/hooks.ts
issmail-basel: @KaWaite , `minWidth` is important for draggable so user can't drag the column less than the defined `minWidth`, by default and in antD example user can, but Red suggested to have this min width.

web/src/components/molecules/Common/ResizableProTable/index.tsx
issmail-basel: @KaWaite , first render this proColumns may be `undefined`, maybe I can check it like this: `if (!proColumns) return`?

web/src/components/molecules/Common/ResizableProTable/index.tsx
issmail-basel: @KaWaite , `width` is a changeable property depending on the dragging, and antD depends on it to set the `width` so I need the minWidth to be a fixed value and prevent the dragging from being less than its value.

web/src/components/molecules/Common/ResizableProTable/index.tsx
issmail-basel: @KaWaite , this condition checks if there is a `columns` definition (not data), so if the user did not specify the columns of the table, I should not show the table at all.
In case we are fetching columns definition from the backend, this condition is to wait for columns before rendering.

web/src/components/molecules/Common/ResizableProTable/resizble.tsx
issmail-basel: @KaWaite , yes, one is changeable the other is fixed and defines the minimum value for narrowing the table.

web/src/components/molecules/Common/ResizableProTable/resizble.tsx
issmail-basel: @KaWaite , fixed.

web/src/components/organisms/Project/Content/ContentList/hooks.ts
KaWaite: Ahhh ok!

web/src/components/molecules/Common/ResizableProTable/index.tsx
KaWaite: I think if it can be undefined, doing `if(!!proColumns)` will suffice

web/src/components/molecules/Common/ResizableProTable/index.tsx
KaWaite: Wouldn't we want to do that kind of checking in the parent component? That way, we can keep this component simpler and show Loading if its loading as well.

web/src/components/molecules/Common/ResizableProTable/index.tsx
issmail-basel: @KaWaite , done. `if(proColumn)` seems to be enough.

web/src/index.css
issmail-basel: @KaWaite , moved to resizable component.

web/src/components/molecules/Common/ResizableProTable/index.tsx
issmail-basel: @KaWaite, totally right.
Changed and I kept the condition in the parent component.

web/src/components/molecules/Common/ResizableProTable/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+import ProTable, {
+  ProColumns,
+  OptionConfig,
+  ListToolBarProps,
+} from "@reearth-cms/components/atoms/ProTable";
+import { ResizableTitle } from "@reearth-cms/components/molecules/Common/ResizableProTable/resizable";
+import type { ResizeCallbackData } from "@reearth-cms/components/molecules/Common/ResizableProTable/resizable";
+
+export type Props = {
+  className?: string;
```
KaWaite, line 13: You've added className here but don't use it.

KaWaite (approved):

PR #327, opened 2022-10-26
fix(web): items form redirection

issmail-basel:
# Changes
- Add item should keep the user on the same page, but now it is editing the saved item
- Edit item should keep the user on the same page as well
- Make sure form data is populated if page is refreshed


web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
```diff
@@ -11,14 +12,23 @@ import { useT } from "@reearth-cms/i18n";
...
+    (modelId: string) => {
+      navigate(`/workspace/${workspaceId}/project/${projectId}/content/${modelId}`);
+    },
+    [navigate, workspaceId, projectId],
+  );
+
+  const handleNavigateBack = useCallback(() => {
+    navigate(`/workspace/${workspaceId}/project/${projectId}/content/${modelId}`);
+  }, [navigate, workspaceId, projectId, modelId]);
+
```
KaWaite, line 31: These are the same so can you just use the handleNavigateToModal for the onBack as well?

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
```diff
@@ -38,9 +48,10 @@ export default ({ itemId }: Props) => {
         Notification.error({ message: t("Failed to create item.") });
         return;
       }
+      navigate(`./${item.data.createItem.item.id}`);
```
KaWaite, line 51: Is this a valid path? If it is, until now we have been writing full paths, so perhaps we should decide which way to go and do that always.

issmail-basel: @KaWaite , the full URL is the same except for the last part, so I thought to use a relative URL to make it short, and avoid unnecessary variables.
Changed to the full path.

web/src/components/organisms/Project/Content/ContentDetails/hooks.ts
issmail-basel: @KaWaite , merged the two into one.

KaWaite (approved):

PR #334, opened 2022-10-27
feat(server): connect thread with asset

nourbalaha:
# Overview
I connected thread with asset

## What I did
- Now When Uploading a new asset, a new thread is being created and saved into the database and then connected to asset via `asset.threadId`
- I made `asset.threadId` mandatory when creating a new asset via asset builder.


server/pkg/asset/builder_test.go
```diff
@@ -122,6 +123,21 @@ func TestBuilder_Build(t *testing.T) {
...
+				project:     pid,
+				createdBy:   uid,
+				fileName:    "hoge",
+				size:        size,
+				previewType: PreviewTypeFromRef(lo.ToPtr(PreviewTypeImage.String())),
+				file:        &f,
+				uuid:        "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
+				thread:      ThreadID{},
+			},
+			err: ErrZeroSize,
```
rot1024, line 139: This test name and `err: ErrZeroSize` is mismatched. Does it test `ErrNoThread`?

server/internal/usecase/interactor/asset.go
```diff
@@ -78,6 +79,14 @@ func (i *Asset) Create(ctx context.Context, inp interfaces.CreateAssetParam, ope
 				return nil, err
 			}
 
+			th, err := thread.New().NewID().Workspace(prj.Workspace()).Comments([]*thread.Comment{}).Build()
```
rot1024, line 82:
```suggestion
			th, err := thread.New().NewID().Workspace(prj.Workspace()).Build()
```

Note that a nil slice and an empty slice is almost same in Golang

rot1024 (approved):

server/pkg/asset/builder_test.go
nourbalaha: This is done by mistake, I pushed a fix. Thanks!

server/internal/usecase/interactor/asset.go
nourbalaha: Done!

PR #336, opened 2022-10-28
fix(server): prevent model's key duplication 

mimoham24:
# Overview


rot1024 (commented):
Add unique index for model key

server/internal/infrastructure/mongo/model.go
```diff
@@ -15,7 +15,7 @@ import (
 
 var (
 	modelIndexes       = []string{"key", "project", "workspace"}
```
rot1024, line 17:
```suggestion
	modelIndexes       = []string{"project", "workspace"}
```

server/pkg/model/model.go
```diff
@@ -8,7 +8,8 @@ import (
 )
 
 var (
-	ErrInvalidKey = errors.New("invalid key")
+	ErrInvalidKey    = errors.New("invalid key")
+	ErrDuplicatedKey = errors.New("duplicated key")
```
rot1024, line 12: Move ErrDuplicatedKey to `internal/usecase/interfaces/model.go`

rot1024 (changes requested):

rot1024 (approved):

PR #342, opened 2022-10-30
feat(web): add model edit/delete

issmail-basel:
# Overview
Add options dropdown to models card which allows to
- Edit model
- Delete model


web/src/components/organisms/Project/Schema/hooks.ts
```diff
@@ -27,8 +26,8 @@ export default () => {
   const [collapsed, collapse] = useState(false);
 
   useEffect(() => {
-    if (!modelId && currentModel) {
-      navigate(`/workspace/${workspaceId}/project/${projectId}/schema/${currentModel.id}`);
+    if (!modelId) {
+      setCurrentModel();
```
KaWaite, line 30: Since you are trying to keep the model setting and updating logic in the ModelsMenu maybe you don't need to be checking for the modelId here since this should be handled by the ModelsMenu, right?
Also, do you think we should be using `useParams` for any of the IDs outside of the Header (for workspace and project IDs) and ModelsMenu (for model ID)??

web/src/components/organisms/Project/ModelsMenu/hooks.ts
```diff
@@ -9,7 +10,7 @@ import {
   Model as GQLModel,
 } from "@reearth-cms/gql/graphql-client-api";
 import { useT } from "@reearth-cms/i18n";
-import { useModel } from "@reearth-cms/state";
+import { useModel, useWorkspace } from "@reearth-cms/state";
```
KaWaite, line 13: Can we use useProject here as well and remove using usParams for projectID? 
Here we only want to get the modelId from the route and rely on the header managing the current workspace and project

web/src/components/molecules/ProjectOverview/ModelDeletionModal/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+      onCancel={() => onClose()}
+      footer={[
+        <Button key="back" onClick={() => onClose()}>
+          {t("Cancel")}
+        </Button>,
+        <Button key="submit" type="primary" onClick={() => onDelete(model?.id)} danger>
+          {t("Delete Model")}
+        </Button>,
+      ]}>
+      <p>{t("Are you sure you want to delete this model ") + model?.name + "?"}</p>
```
KaWaite, line 30: `{`${t("Are you sure you want to delete this model")} (${model?.name})?`}`

KaWaite, line 30: The final should be:

Are you sure you want to delete the model **SomeModel123**?

You might need to wrap the model?.name in a span and change the styles for it to be bolded (like the design).

web/src/components/molecules/ProjectOverview/ModelDeletionModal/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <Alert
+        message={t("Warning")}
+        description={t("This Change would be permanent and can not be rolled back.")}
```
KaWaite, line 33: This action will permanently delete the selected model and cannot be reversed.

web/src/components/molecules/ProjectOverview/ModelCard.tsx
```diff
@@ -1,28 +1,59 @@
...
+    <Menu
+      items={[
+        {
+          key: "edit",
+          label: "Edit",
+          onClick: () => onModelUpdateModalOpen(model),
+        },
+        {
+          key: "delete",
+          label: "Delete",
```
KaWaite, line 42: Both labels should be wrapped with i18n

web/src/components/molecules/ProjectOverview/ModelDeletionModal/index.tsx
issmail-basel: @KaWaite , changed it to bold.

web/src/components/molecules/ProjectOverview/ModelDeletionModal/index.tsx
issmail-basel: @KaWaite , done.

web/src/components/molecules/ProjectOverview/ModelCard.tsx
issmail-basel: @KaWaite , wrapped the labels with `i18n`

web/src/components/organisms/Project/Schema/hooks.ts
issmail-basel: @KaWaite , done.

I think so, so we can keep one place for updating their stated, and also one instance selected as a time, also it will easier for developing because developer can use the state directly without thinking much about the way it is fetched.

web/src/components/organisms/Project/ModelsMenu/hooks.ts
issmail-basel: @KaWaite , also updated.
https://github.com/reearth/reearth-cms/blob/a92e02ae442c0515eece917d5004c0ad94bd662c/web/src/components/organisms/Project/ModelsMenu/hooks.ts#L13

web/src/components/molecules/ProjectOverview/ModelDeletionModal/index.tsx
KaWaite: A you missed a small change. Not `...this model Blahblah` but `...the model Blahblah`

KaWaite (approved):

PR #346, opened 2022-10-31
feat(web): implement comments count column in asset list table

nourbalaha:
# Overview
In this PR I implements the comments count column in asset list table

## What I have done
- I added thread to asset schema.
- I added Comment type to asset.type.ts.
- I added convertComment to convertAsset.ts.
- I implemented comments count in asset list table.


web/src/components/molecules/Asset/AssetListTable/index.tsx
```diff
@@ -56,7 +56,10 @@ const AssetListTable: React.FC<AssetListTableProps> = ({
       title: () => <Icon icon="message" />,
       dataIndex: "commentsCount",
       key: "commentsCount",
-      render: (_, _asset) => <CustomTag value={0} />,
+      render: (_, asset) => {
+        const commentCount = asset.comments?.length || 0;
+        return <CustomTag value={commentCount} />;
```
KaWaite, line 61: To keep this simple and concise just do 
```
<CustomTag value={asset.comments?.length || 0} />
```

web/src/components/organisms/Asset/convertAsset.ts
```diff
@@ -13,5 +18,16 @@ export const convertAsset = (GQLAsset: GQLAsset | undefined): Asset | undefined
     projectId: GQLAsset.projectId,
     size: GQLAsset.size,
     url: GQLAsset.url,
+    threadId: GQLAsset.thread?.id ?? "",
+    comments: GQLAsset.thread?.comments?.map(comment => convertComment(comment)) ?? [],
+  };
+};
+
+export const convertComment = (GQLComment: GQLComment): Comment => {
```
KaWaite, line 26: For now its fine, but notice we have I think a few convert functions now for converting to/from GQL types. 
It'd be nice to consolidate them in a single file (like we do in reearth-web).

nourbalaha: Ok, this makes sense. I will do it in a separate PR.

web/src/components/molecules/Asset/AssetListTable/index.tsx
nourbalaha: Done!

KaWaite (approved):

PR #350, opened 2022-10-31
feat(web): add key to fields cards

issmail-basel:
# Overview
- add key to fields cards


web/src/components/molecules/Schema/ModelFieldList.tsx
```diff
@@ -48,7 +48,12 @@ const ModelFieldList: React.FC<Props> = ({
                 <h3>{(item as Field).type}</h3>
               </FieldThumbnail>
             }
-            title={`${(item as Field).title}${(item as Field).required ? " *" : ""}`}
+            title={
```
KaWaite, line 51: Is `title` suppose to be a string or component? We also don't know how the title is handled (wrapped in a div or not), do we? It seems weird to wrap the key in a span but do nothing to the title.

issmail-basel: @KaWaite 
https://github.com/reearth/reearth-cms/blob/a536b07e7cabf172616429a0a57fd60b19f8b735/web/src/components/molecules/Schema/ModelFieldList.tsx#L51-58

KaWaite (approved):

PR #353, opened 2022-11-01
feat(web): disable button on validation

issmail-basel:
## fixes
- [ ] FE: When user hasn’t finished filling fields, “OK” button should be disabled for items


web/src/components/molecules/Schema/FieldModal/FieldUpdateModal/index.tsx
```diff
@@ -58,6 +58,7 @@ const FieldUpdateModal: React.FC<Props> = ({
 }) => {
   const t = useT();
   const [form] = Form.useForm();
+  const [buttonDisabled, setButtonDisabled] = useState(false);
```
KaWaite, line 61: This should be true, right?

issmail-basel: @KaWaite , `false` because when we open the update modal, the data inside it should already pass the validation.

KaWaite (approved):

PR #354, opened 2022-11-01
feat(web): disable model form button on validation

issmail-basel:
## Fixes
- [ ] FE: When user hasn’t finished filling fields, “OK” button should be disabled for items


web/src/components/molecules/Schema/ModelCreationModal/index.tsx
```diff
@@ -57,8 +58,27 @@ const ModelCreationModal: React.FC<Props> = ({
...
+      okButtonProps={{ disabled: buttonDisabled }}>
+      <Form
+        form={form}
+        layout="vertical"
+        initialValues={initialValues}
+        onValuesChange={() => {
+          form
+            .validateFields()
+            .then(() => {
+              setButtonDisabled(true);
```
KaWaite, line 74: This should be setting to false

KaWaite (approved):

PR #356, opened 2022-11-01
fix: integrations table

issmail-basel:
## Fixes
- Clear selected integration when submitting integration to connect
- Show the right value for the role when editing an integration, the value wasn't synced


web/src/components/organisms/Settings/Integration/hooks.ts
```diff
@@ -62,7 +64,8 @@ export default (workspaceId?: string) => {
         (integrationMember): integrationMember is IntegrationMember =>
           !!integrationMember &&
           integrationMember.integration.name.toLowerCase().includes(searchTerm ?? ""),
-      );
+      )
+      .sort((integration1, integration2) => stringSortCallback(integration1.id, integration2.id));
```
KaWaite, line 68: I'm really confused about this code. The variable is called workspaceIntegrationMembers, but it seems like we are using this for the Integrations list, which is a list of integrations and doesn't seem to show members other than the creator.
Can you either fix the naming so its clear what it is we are doing here? Or if it makes sense to you, perhaps my understanding of integrations is limited and I need you or @lavalse to explain the system to me more...

issmail-basel: @KaWaite , the integrations are treated as members, in CMS we have two types of members, either:
- User member
- Integration member
so the member is one of two possible types: userMembers or integrationMembers.

KaWaite (approved):

PR #363, opened 2022-11-03
fix(server): save unique, multiValue, and required on update

mimoham24:
# Overview

### DIFF
--- a/server/internal/usecase/interactor/schema.go
+++ b/server/internal/usecase/interactor/schema.go
@@ -150,9 +150,17 @@ func (i Schema) UpdateField(ctx context.Context, param interfaces.UpdateFieldPar
 				}
 				f.SetKey(k)
 			}
-			f.SetRequired(*param.Required)
-			f.SetUnique(*param.Unique)
-			f.SetMultiValue(*param.MultiValue)
+			if param.Required != nil {
+				f.SetRequired(*param.Required)
+			}
+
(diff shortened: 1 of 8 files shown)


server/internal/usecase/interactor/schema.go
```diff
@@ -150,6 +150,9 @@ func (i Schema) UpdateField(ctx context.Context, param interfaces.UpdateFieldPar
 				}
 				f.SetKey(k)
 			}
+			f.SetRequired(*param.Required)
+			f.SetUnique(*param.Unique)
+			f.SetMultiValue(*param.MultiValue)
```
rot1024, line 155:
```suggestion
			if param.Required != nil {
				f.SetRequired(*param.Required)
			}

			if param.Unique != nil {
				f.SetUnique(*param.Unique)
			}

			if param.MultiValue != nil {
				f.SetMultiValue(*param.MultiValue)
			}
```

Have you checked that it works properly? It seems to me that some parameters cause panic.

rot1024 (changes requested):

server/internal/usecase/interactor/schema.go
mimoham24: after a meeting with Basel we agreed to send the parameters again on update. however, those checks are safer. thank you

rot1024 (approved):

PR #364, opened 2022-11-03
fix(web): contents overflows layout

issmail-basel:
# Overview
When we have a large number of schema fields, content overflows the layout:

and

After the fix:

### DIFF
--- a/web/src/components/atoms/InnerContents/complex.tsx
+++ b/web/src/components/atoms/InnerContents/complex.tsx
@@ -24,7 +24,7 @@ export default ComplexInnerContents;
 const PaddedContent = styled.div`
   display: flex;
   margin: 16px 0 0 16px;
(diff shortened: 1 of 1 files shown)


web/src/components/atoms/InnerContents/complex.tsx
```diff
@@ -24,7 +24,7 @@ export default ComplexInnerContents;
 const PaddedContent = styled.div`
   display: flex;
   margin: 16px 0 0 16px;
-  height: calc(100vh - 66px);
+  min-height: calc(100vh - 66px);
```
KaWaite, line 27: Please keep it height, and add overflow: auto; to the PaddContent. If you check test environment, if you make the height short, you'll notice the right fields list and the list of models scroll by themselves. With your fix here it breaks that and EVERYTHING scrolls together, which isn't great.

PR #369, opened 2022-11-04
feat(web): hide subtitle if does not exist

issmail-basel:
# Overview
When there is no key, do not show anything instead of `undefined`


web/src/components/molecules/Content/List/index.tsx
```diff
@@ -37,7 +37,7 @@ const ContentListMolecule: React.FC<Props> = ({
     <ContentWrapper modelsMenu={ModelsMenu}>
       <PageHeader
         title={model?.name}
-        subTitle={`#${model?.key}`}
+        subTitle={model?.key ? "#" + model.key : null}
```
KaWaite, line 40:
```suggestion
        subTitle={model?.key ? `#${model.key}` : null}
```

web/src/components/molecules/Schema/index.tsx
```diff
@@ -42,7 +42,7 @@ const Schema: React.FC<Props> = ({
       }
       center={
         <Content>
-          <PageHeader title={model?.name} subTitle={`#${model?.key}`} />
+          <PageHeader title={model?.name} subTitle={model?.key ? "#" + model.key : null} />
```
KaWaite, line 45:
```suggestion
          <PageHeader title={model?.name} subTitle={model?.key ? `#${model.key}` : null} />
```

web/src/components/molecules/Content/List/index.tsx
issmail-basel: @KaWaite , updated.

web/src/components/molecules/Schema/index.tsx
issmail-basel: @KaWaite , updated.

KaWaite (approved):

PR #376, opened 2022-11-07
feat(server): add debug user to auth middleware

yk-eukarya:
# Overview
Fix for graphQL playground, and add debug users support


server/internal/app/app.go
```diff
@@ -52,7 +52,7 @@ func initEcho(ctx context.Context, cfg *ServerConfig) *echo.Echo {
 	}
 
 	internalJWTMiddleware := echo.WrapMiddleware(lo.Must(
-		appx.AuthMiddleware(cfg.Config.JWTProviders(), adapter.ContextAuthInfo, false),
+		appx.AuthMiddleware(cfg.Config.JWTProviders(), adapter.ContextAuthInfo, cfg.Debug || cfg.Config.Dev),
```
rot1024, line 55:
```suggestion
		appx.AuthMiddleware(cfg.Config.JWTProviders(), adapter.ContextAuthInfo, true),
```

server/internal/app/auth_client.go
```diff
@@ -39,7 +40,20 @@ func authMiddleware(cfg *ServerConfig) echo.MiddlewareFunc {
...
+			if val := req.Header.Get(debugUserHeaderKey); cfg.Debug && val != "" {
+				uId, err := id.UserIDFrom(val)
+				if err != nil {
+					return err
+				}
+				u, err = cfg.Repos.User.FindByID(ctx, uId)
+				if err != nil {
+					return err
+				}
+			}
```
rot1024, line 54:
```suggestion
			if cfg.Debug {
				if val := req.Header.Get(debugUserHeaderKey); val != "" {
					uId, err := id.UserIDFrom(val)
					if err != nil {
						return err
					}
					u, err = cfg.Repos.User.FindByID(ctx, uId)
					if err != nil {
						return err
					}
				}
			}
```

server/internal/app/auth_client.go
```diff
@@ -49,16 +63,27 @@ func authMiddleware(cfg *ServerConfig) echo.MiddlewareFunc {
...
+			if val := req.Header.Get(debugIntegrationHeaderKey); cfg.Debug && val != "" {
+				iId, err := id.IntegrationIDFrom(val)
+				if err != nil {
+					return err
+				}
+				i, err = cfg.Repos.Integration.FindByID(ctx, iId)
+				if err != nil {
+					return err
+				}
+			}
```
rot1024, line 84:
```suggestion
			if cfg.Debug {
				if val := req.Header.Get(debugIntegrationHeaderKey); val != "" {
					iId, err := id.IntegrationIDFrom(val)
					if err != nil {
						return err
					}
					i, err = cfg.Repos.Integration.FindByID(ctx, iId)
					if err != nil {
						return err
					}
				}
			}
```

rot1024 (changes requested):

rot1024 (approved):

PR #378, opened 2022-11-08
fix(worker): not return error when extension isn't supported

HideBa:
# Why
* Since CloudTasks retries until Worker returns 2xx response. I change this.


worker/internal/usecase/interactor/decompress.go
```diff
@@ -28,6 +30,10 @@ func (u *Usecase) Decompress(ctx context.Context, assetID, assetPath string) err
 
 	de, err := decompressor.New(compressedFile, size, ext, uploadFunc)
 	if err != nil {
+		if errors.Is(err, decompressor.ErrUnsupportedExtention) {
+			log.Infof("unsupported extension: decompression skipped AssetID=%s Ext=%s", assetID, ext)
```
rot1024, line 34:
```suggestion
			log.Infof("unsupported extension: decompression skipped assetID=%s ext=%s", assetID, ext)
```

rot1024 (approved):

PR #385, opened 2022-11-08
fix(web): wrong validation error when removing an option

issmail-basel:
## Fixes
- FE: when an option isn’t the last index, it returns an error

## How to reproduce
- In the schema page, click on create a new `options` field.
- Add multiple options for this field
- If you remove the first option, a validation error will appear even if there are no validation errors
Same for updating an `options` field

## Screenshots

After removing the first option `Option1` the following error appears


web/src/components/molecules/Schema/FieldModal/FieldCreationModal/index.tsx
```diff
@@ -170,16 +170,20 @@ const FieldCreationModal: React.FC<Props> = ({
...
-            .then(() => {
-              setButtonDisabled(false);
-            })
-            .catch(fieldsError => {
-              setButtonDisabled(
-                fieldsError.errorFields.some((item: FieldError) => item.errors.length > 0),
-              );
-            });
+          setTimeout(() => {
+            console.log(form.getFieldsValue());
```
KaWaite, line 174: Don't need

web/src/components/molecules/Schema/FieldModal/FieldUpdateModal/index.tsx
```diff
@@ -195,16 +195,20 @@ const FieldUpdateModal: React.FC<Props> = ({
...
-            .then(() => {
-              setButtonDisabled(false);
-            })
-            .catch(fieldsError => {
-              setButtonDisabled(
-                fieldsError.errorFields.some((item: FieldError) => item.errors.length > 0),
-              );
-            });
+          setTimeout(() => {
+            console.log(form.getFieldsValue());
```
KaWaite, line 199: Don't need

web/src/components/molecules/Schema/FieldModal/FieldCreationModal/index.tsx
```diff
@@ -170,16 +170,20 @@ const FieldCreationModal: React.FC<Props> = ({
...
+
+            form
+              .validateFields()
+              .then(() => {
+                setButtonDisabled(false);
+              })
+              .catch(fieldsError => {
+                setButtonDisabled(
+                  fieldsError.errorFields.some((item: FieldError) => item.errors.length > 0),
+                );
```
KaWaite, line 184: Do we really need to make this check? If we are in the catch, isn't there a problem?

issmail-basel: @KaWaite , that's the way validation works in `antd`, it sends the validation errors in the catch statement of `validate` method.

web/src/components/molecules/Schema/FieldModal/FieldCreationModal/index.tsx
issmail-basel: @KaWaite , removed.

web/src/components/molecules/Schema/FieldModal/FieldUpdateModal/index.tsx
issmail-basel: @KaWaite , removed.

web/src/components/molecules/Schema/FieldModal/FieldCreationModal/index.tsx
KaWaite: So is there a chance we will get to the catch even if there are no errors?

issmail-basel: @KaWaite , yes anytime any field has a validation missing, it throws an error to the `catch`, so the first block always runs if there are no validation issues, if there is any input with a validation issue it throws it in `catch`.
the checking `fieldsError.errorFields.some` in case any other issue occurs other than the validation issue, as it contains other properties as well, and it may reach the `catch` statement.

KaWaite (approved):

PR #387, opened 2022-11-08
feat(web): add members search

issmail-basel:
## Fixes
- FE: search function should work


web/src/components/organisms/Settings/Members/hooks.ts
```diff
@@ -88,9 +93,11 @@ export default ({ workspaceId }: Props) => {
             }
           : undefined,
       )
-      .filter((user): user is Member => !!user)
+      .filter(
+        (user): user is Member => !!user && user.user.name.toLowerCase().includes(searchTerm ?? ""),
```
KaWaite, line 97:
```suggestion
        (user): user is Member => !!user && user.user.name.toLowerCase().includes(searchTerm.toLowerCase() ?? ""),
```

KaWaite: Do we want to make the search case-sensitive?

issmail-basel: @KaWaite , no it should be insensitive.
Fixed.

KaWaite (approved):

PR #390, opened 2022-11-08
feat(web): asset updates for integration api

nourbalaha:
# Overview
In this PR I've updated asset for the new integration api schema changes.

## What I have done
1- Asset.CreatedBy is now a union of two types "User | Integration".
2- Asset.CreatedByType is added.
3- CreateAsset mutation does not need createdById as an argument from the front end anymore, it will be added from the operator in the backend.


web/src/components/organisms/Asset/AssetList/hooks.ts
```diff
@@ -90,11 +82,11 @@ export default (projectId?: string) => {
   const createAssets = useCallback(
     (files: UploadFile<File>[]) =>
       (async () => {
-        if (!projectId || !currentUser?.id) return;
+        if (!projectId) return;
         const results = await Promise.all(
           files.map(async file => {
             const result = await createAssetMutation({
-              variables: { projectId, createdById: currentUser?.id, file },
+              variables: { projectId, file },
```
KaWaite, line 89: By removing the createdById, how do we know who created it? (Especially now that we have createdBy and createdByType, don't we need to send created by Id AND the type?

nourbalaha: There is no need to pass `createdById` as an argument from the front end when uploading a new asset any more.
createdById is being added from the operator in the backend instead.

KaWaite (approved):

PR #392, opened 2022-11-09
feat(server): add machine operator

HideBa:


rot1024 (changes requested):
- It is not up to the use case layer to decide if it is a CMS Operator or not, but at the point of the app middleware that generates the Operator. Based on the credentials, determine if it is a User, Integration, or CMS. In the use case layer, simply add a conditional branch to the EventOperator method if it is a CMS operator.
- The name "CMS" is not a good fit. "Machine" is a better name for a specific application.

server/internal/usecase/operator.go
```diff
@@ -11,6 +11,7 @@ import (
 type Operator struct {
 	User               *user.ID
 	Integration        *integration.ID
+	Machine            *bool
```
rot1024, line 14:
```suggestion
	Machine            bool
```

server/internal/usecase/operator.go
```diff
@@ -115,5 +116,8 @@ func (o *Operator) EventOperator() event.Operator {
 	if o.Integration != nil {
 		eOp = event.OperatorFromIntegration(*o.Integration)
 	}
+	if o.Machine != nil {
```
rot1024, line 119:
```suggestion
	if o.Machine {
```

server/pkg/event/operator.go
```diff
@@ -32,14 +32,14 @@ func (o Operator) Integration() *IntegrationID {
...
 
-func (o Operator) CMS() bool {
-	return o.isCMS
+func (o Operator) Machine() bool {
+	return o.isMachine
 }
 
-func (o Operator) CMSRef() *bool {
-	return &o.isCMS
+func (o Operator) MachineRef() *bool {
```
rot1024, line 39:
```suggestion
func (o Operator) Machine() bool {
```

server/internal/infrastructure/mongo/mongodoc/event.go
```diff
@@ -29,7 +29,7 @@ func NewEvent(e *event.Event[any]) (*EventDocument, string, error) {
 		Timestamp:   e.Timestamp(),
 		User:        e.Operator().User().StringRef(),
 		Integration: e.Operator().Integration().StringRef(),
-		CMS:         e.Operator().CMSRef(),
+		CMS:         e.Operator().MachineRef(),
```
rot1024, line 32:
```suggestion
		Machine:         e.Operator().Machine(),
```

rot1024 (changes requested):

rot1024 (approved):

PR #395, opened 2022-11-10
feat(web): add multiple users to workspaces

nourbalaha:
# Overview
In this PR I supported adding multiple users to workspaces


web/src/components/molecules/Member/MemberAddModal/index.tsx
```diff
@@ -7,47 +7,42 @@ import Form from "@reearth-cms/components/atoms/Form";
...
-          id: string;
-          name: string;
-          email: string;
-        }
-      | undefined
-    >
-  >;
+  onSubmit?: (users: MemberInput[]) => void;
+  searchedUser: User | undefined;
+  changeSearchedUser: React.Dispatch<React.SetStateAction<User | undefined>>;
```
KaWaite, line 26: I think for useState setters you can type it as () => void;

web/src/components/molecules/Member/MemberAddModal/index.tsx
```diff
@@ -61,24 +56,37 @@ const MemberAddModal: React.FC<Props> = ({
...
   );
 
-  const handleMemberRemove = useCallback(() => {
-    form.resetFields();
-    changeSearchedUser(undefined);
-  }, [changeSearchedUser, form]);
+  const handleMemberRemove = useCallback(
+    (userId: string) => {
+      const list = searchedUserList.filter(user => user.id !== userId);
+      changeSearchedUserList(list);
```
KaWaite, line 62: This should be all you need 
```
changeSearchedUserList(oldList => {
  return oldList.filter(user => user.id !== userId
})
```

web/src/components/organisms/Settings/Members/hooks.ts
```diff
@@ -82,6 +81,16 @@ export default ({ workspaceId }: Props) => {
     [searchUserQuery],
   );
 
+  const handleUserAdd = useCallback(() => {
+    if (
+      searchedUser &&
+      searchedUser?.id !== data?.me?.id &&
```
KaWaite, line 87:
```suggestion
      searchedUser.id !== data?.me?.id &&
```

web/src/components/molecules/Workspace/types.ts
```diff
@@ -18,6 +18,11 @@ export type Member = {
   };
 };
 
+export type MemberInput = {
```
KaWaite, line 21: Can this be UserInput?

KaWaite: Or just User

web/src/components/organisms/Settings/Members/hooks.ts
```diff
@@ -100,30 +109,28 @@ export default ({ workspaceId }: Props) => {
...
+        const result = await addUsersToWorkspaceMutation({
+          variables: { workspaceId, users: users as GQLMemberInput[] },
+          refetchQueries: ["GetWorkspaces"],
+        });
+        const workspace = result.data?.addUsersToWorkspace?.workspace;
+        if (result.errors || !workspace) {
+          Notification.error({ message: t("Failed to add one or more members.") });
+          return;
+        }
+        setWorkspace(workspace);
```
KaWaite, line 127: We handle workspace changing (and project changing) basically exclusively in the Header now, so please remove this.
Anyways, it makes no sense to change the workspace on user add.

web/src/components/organisms/Settings/Members/hooks.ts
```diff
@@ -100,30 +109,28 @@ export default ({ workspaceId }: Props) => {
...
+
+        if (result) {
```
KaWaite, line 129: I think you can check like `if (result.workspace)` or something to be a little more specific, no?

web/src/components/organisms/Settings/Members/index.tsx
```diff
@@ -154,10 +157,13 @@ const Members: React.FC = () => {
       <MemberAddModal
         open={MemberAddModalShown}
         searchedUser={searchedUser}
+        searchedUserList={searchedUserList}
+        changeSearchedUserList={changeSearchedUserList}
         onClose={handleMemberAddModalClose}
         handleUserSearch={handleUserSearch}
+        handleUserAdd={handleUserAdd}
```
KaWaite, line 164: `onUserAdd`
Also, please update the prop above it too, handleUserSearch => onUserSearch

web/src/components/molecules/Member/MemberAddModal/index.tsx
nourbalaha: Done!
I did this for `changeSearchedUser` 
however for `changeSearchedUserList: React.Dispatch<React.SetStateAction<User[]>>`
I kept it the way it was because of this code:
```js
      changeSearchedUserList((oldList: User[]) =>
        oldList.filter((user: User) => user.id !== userId),
      );
```

web/src/components/organisms/Settings/Members/hooks.ts
nourbalaha: Done!

web/src/components/organisms/Settings/Members/hooks.ts
nourbalaha: Done!
I did not add this, it was in the original code, I removed it anyways.

web/src/components/organisms/Settings/Members/hooks.ts
nourbalaha: How about `if (result.data)`?

web/src/components/organisms/Settings/Members/index.tsx
nourbalaha: Done!

web/src/components/molecules/Workspace/types.ts
KaWaite: OK

web/src/components/organisms/Settings/Members/hooks.ts
KaWaite: Sorry I know it is a little out of scope for this PR, but can you also remove all use of `setWorkspace` from this file and see if everything still works ok? 
Also, can you remove the passed workspaceId prop (and its useParam in the index.tsx) and use the `currentWorkspace.id` instead?

web/src/components/organisms/Settings/Members/hooks.ts
KaWaite: That! haha Because if my memory is correct if there is an error there won't be result.data.

web/src/components/organisms/Settings/Members/hooks.ts
nourbalaha: After doing this refetching does not work anymore. somehow using setWorkspace is triggering refetching the table members. can we handle this in a separate PR?

I used `currentWorkspace.id` instead of `workspaceId`

web/src/components/organisms/Settings/Members/hooks.ts
nourbalaha: So which one should it be?
result.data
result.context
result.extensions

web/src/components/organisms/Settings/Members/hooks.ts
KaWaite: Yeah that isn't a good way to handle the refetch... Can you make sure you handle this in a different PR? You can close this convo after you read this.

web/src/components/organisms/Settings/Members/hooks.ts
KaWaite: How you updated it is good I think (result.data)!

KaWaite (approved):

web/src/components/organisms/Settings/Members/hooks.ts
nourbalaha: Sure, I will handle this in a different PR, I don't think that it will be soon though.

PR #404, opened 2022-11-10
feat(server): integration API to update item

yk-eukarya:
# Overview
Add update item API to integration APIs


rot1024 (approved):
There are a few problems. But we can fix them later.

It is a waste of communication to retrieve all versions from the database when only the latest version is used.

PR #420, opened 2022-11-14
fix(server): integration bugs

yk-eukarya:


server/internal/adapter/integration/item_api.go
```diff
@@ -87,7 +87,7 @@ func (s Server) ItemCreate(ctx context.Context, request ItemCreateRequestObject)
 		return nil, err
 	}
 
-	return ItemCreate200JSONResponse(toItem(i, ver[len(ver)-1], id.NewModelID())), nil
+	return ItemCreate200JSONResponse(toItem(i, ver[len(ver)-1], mId)), nil
```
rot1024, line 90: please fix `toItem` func to use item's modelID without model ID from outside: `item.Model()`

rot1024 (approved):

PR #428, opened 2022-11-15
feat(web): asset unlinking and deletion functionality

KaWaite:
# Overview


web/src/components/atoms/DownloadButton/index.tsx
```diff
@@ -5,37 +5,69 @@ import Button, { ButtonProps } from "@reearth-cms/components/atoms/Button";
...
+  fileName: string;
+  previewType?: PreviewType;
+  projectId: string;
+  size: number;
+  url: string;
+  threadId: string;
+  comments: Comment[];
+};
+
+export type Comment = {
```
issmail-basel, line 33: Atom is the smallest component that is aware of itself, and should not know about anything outside its scope.
I'm not sure if the `comments` and `Assets` declarations are right to be here, as they represent things outside the `DownloadButton` scope.

web/src/components/atoms/DownloadButton/index.tsx
```diff
@@ -5,37 +5,69 @@ import Button, { ButtonProps } from "@reearth-cms/components/atoms/Button";
...
+  author: string;
+  content: string;
+  createdAt: string;
+};
+
 type DownloadButtonProps = {
   title?: string;
-  filename: string;
-  url: string;
+  selected?: Asset[];
```
issmail-basel, line 42: Can't we generalize that, and maybe accept other types, it's not mandatory to have an `asset` to make this atom work, right?
What do you think?

web/src/components/atoms/Icon/Icons/clear.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
issmail-basel, line 1: I think no need to add the SVG file.
The same icon exists in `antd` icons its name `ClearOutlined`

web/src/components/atoms/Icon/icons.ts
```diff
@@ -79,6 +82,8 @@ export default {
   listBullets: ListBullets,
   numberNine: NumberNine,
   link: Link,
+  linkSolid: LinkSolid,
+  unlinkSolid: Unlink,
```
issmail-basel, line 86: Maybe it's better to name the second one `UnlinkSolid`.

web/src/components/atoms/DownloadButton/index.tsx
```diff
@@ -5,37 +5,69 @@ import Button, { ButtonProps } from "@reearth-cms/components/atoms/Button";
...
-    fileDownload(blob, filename);
-  }, [filename, url]);
+  }, [selected]);
 
   return (
     <Button
       icon={displayDefaultIcon && <Icon icon="download" />}
       onClick={handleDownload}
-      disabled={!url}
+      disabled={!selected}
```
issmail-basel, line 70: What if we passed an empty array as `selected` (the condition will give false and the button will be enabled), this will fail I think.

web/src/components/molecules/Asset/Asset/AssetBody/Asset.tsx
```diff
@@ -119,7 +119,7 @@ const AssetMolecule: React.FC<Props> = ({
             />
           </Card>
         )}
-        <DownloadButton type="ghost" filename={asset.fileName} url={asset.url} displayDefaultIcon />
+        <DownloadButton type="ghost" selected={asset ? [asset] : undefined} displayDefaultIcon />
```
issmail-basel, line 122: I prefer to have this check inside the atom, so it will give a higher flexibility, to either pass an array or a single file.

web/src/components/molecules/Asset/AssetListTable/index.tsx
```diff
@@ -140,10 +133,27 @@ const AssetListTable: React.FC<AssetListTableProps> = ({
     },
   };
 
+  const AlertOptions = (props: any) => {
+    return (
+      <Space size={16}>
+        <a
+          onClick={props.onCleanSelected}
+          style={{ display: "flex", alignItems: "center", gap: "8px" }}>
```
issmail-basel, line 141: a lot of inline styles in nested elements, wouldn't be cleaner if used styled-components for these styles?

web/src/components/molecules/Common/LinkAssetModal/LinkAssetModal.tsx
```diff
@@ -49,6 +53,7 @@ const LinkAssetModal: React.FC<Props> = ({
   handleUpload,
 }) => {
   const t = useT();
+  const [hovered, setHovered] = useState<string>();
```
issmail-basel, line 56: Maybe `hoveredId` instead of `hovered` (same for set), hovered alone does not describe this variable's contents

web/src/components/atoms/DownloadButton/index.tsx
KaWaite: Yeah, I was just trying to get the asset table UI/UX updated and merged before the demo so was ignoring some things I didn't want to and this is one of them. 
Not sure the DownloadButton is really necessary/an atom. For refactoring I'll want to maybe move this to be a molecule, stop using it in the asset table all together, make it generic, or something else lol

web/src/components/atoms/DownloadButton/index.tsx
KaWaite: Oh you are ahead of me. Like I mentioned in the reply above, making this component generic is definitely an option. Though, if we make it generic we will pull basically everything out of it that makes it unique to an Antd Button, no? 
Anyways, it is only being used with assets (schema and content pages). 
I think making it a common molecule as AssetDownloadButton or something is the best option.

web/src/components/atoms/Icon/Icons/clear.svg
KaWaite: Oh didn't see that! Thanks

web/src/components/atoms/Icon/icons.ts
KaWaite: Changed!

web/src/components/atoms/DownloadButton/index.tsx
KaWaite: Changed to `!selected || selected.length <= 0` to account for empty arrays (though where this component is being used sets selected to undefined

web/src/components/molecules/Asset/Asset/AssetBody/Asset.tsx
KaWaite: I'll leave the refactoring/removing/etc of the Download button up to you, okay? It isn't a great component right now and again would make more sense as a molecule or to be removed completely since it is just a glorified Antd Button.

web/src/components/molecules/Asset/AssetListTable/index.tsx
KaWaite: Done

web/src/components/molecules/Common/LinkAssetModal/LinkAssetModal.tsx
KaWaite: True. Changed to `hoveredAssetId` to be as clear as possible

web/src/components/molecules/Asset/Asset/AssetBody/Asset.tsx
issmail-basel: @KaWaite , ok, sure.

issmail-basel: I'll resolve this for now.

web/src/components/atoms/DownloadButton/index.tsx
issmail-basel: @KaWaite , ok, will keep this in mind :+1: .

PR #430, opened 2022-11-15
feat(server): add thread to the item

mimoham24:


server/internal/usecase/interactor/item.go
```diff
@@ -96,6 +96,7 @@ func (i Item) Create(ctx context.Context, param interfaces.CreateItemParam, oper
 			Schema(param.SchemaID).
 			Project(s.Project()).
 			Model(param.ModelID).
+			Thread(id.NewThreadID()).
```
rot1024, line 99: Is a thread actually initialized? An ID is created, but model seems not.

server/pkg/item/builder.go
```diff
@@ -74,6 +77,11 @@ func (b *Builder) Project(pid ProjectID) *Builder {
 	return b
 }
 
+func (b *Builder) Thread(tid ThreadID) *Builder {
+	b.i.thread = tid
+	return b
+}
+
```
rot1024, line 84:
```suggestion
func (b *Builder) Thread(tid ThreadID) *Builder {
	b.i.thread = tid
	return b
}

func (b *Builder) NewThread() *Builder {
	b.i.thread = NewThreadID()
	return b
}

```

NewThread method may be helpful

server/schemas/item.graphql
```diff
@@ -1,10 +1,12 @@
 type Item implements Node {
   id: ID!
   schemaId: ID!
+  threadId: ID!
   modelId: ID!
   projectId: ID!
-  project: Project!
   schema: Schema!
+  project: Project!
```
rot1024, line 8:
```graphql
  schema: Schema
  project: Project
```

Schema and project can be optional if you make thread optional.

rot1024 (changes requested):

PR #434, opened 2022-11-17
feat(web): implement upload asset from url

nourbalaha:
# Overview
In this PR I implemented upload asset from url in asset list, schema and item pages.


web/src/components/organisms/Asset/AssetList/hooks.ts
```diff
@@ -107,6 +110,26 @@ export default () => {
...
+        if (!projectId) return;
+        const result = await createAssetMutation({
+          variables: { projectId, file: null, url },
+        });
+        if (result.errors || !result.data?.createAsset) {
+          Notification.error({ message: t("Failed to add asset.") });
+          return;
+        }
+        if (result.data?.createAsset) {
+          Notification.success({ message: t("Asset successfully added!") });
```
KaWaite, line 125: Successfully added asset!

web/src/components/organisms/Asset/AssetList/hooks.ts
```diff
@@ -154,6 +177,14 @@ export default () => {
     navigate(`/workspace/${workspaceId}/project/${projectId}/asset/${asset.id}`);
   };
 
+  const hideUploadModal = useCallback(() => {
```
KaWaite, line 180: handleUploadModalClose

web/src/components/molecules/Asset/AssetList/index.tsx
```diff
@@ -19,7 +21,13 @@ type Props = {
   uploading: boolean;
   uploadModalVisibility: boolean;
   loading: boolean;
-  onAssetCreate: (files: UploadFile[]) => Promise<(Asset | undefined)[]>;
+  uploadUrl: string;
+  uploadType: UploadType;
+  hideUploadModal: () => void;
```
KaWaite, line 26: onUploadModalClose

web/src/components/molecules/Schema/FieldModal/FieldCreationModal/index.tsx
```diff
@@ -40,7 +41,13 @@ export type Props = {
   loadingAssets: boolean;
   uploading: boolean;
   uploadModalVisibility: boolean;
-  createAssets: (files: UploadFile[]) => Promise<(Asset | undefined)[]>;
+  uploadUrl: string;
+  uploadType: UploadType;
+  hideUploadModal: () => void;
```
KaWaite, line 46: onUploadModalClose

web/src/components/molecules/Common/Form/AssetItem/hooks.ts
```diff
@@ -24,25 +28,42 @@ export default (
...
+        default:
+          assets = await onAssetsCreate(fileList);
+          if (assets.length > 0) asset = assets[0];
+          break;
+      }
+      if (asset) onLink(asset);
+      hideUploadModal();
+    } catch (error) {
+      hideUploadModal();
+    }
```
KaWaite, line 50: You aren't handling this error. Also, if all routes end up with hideUploadModal, put it outside of any conditional statements at the end of the callback.

web/src/components/molecules/Common/Form/AssetItem/hooks.ts
```diff
@@ -24,25 +28,42 @@ export default (
...
-  }, [setFileList, setUploadModalVisibility, setUploading]);
-
   const handleUpload = useCallback(async () => {
     setUploading(true);
-    const assets = await createAssets(fileList);
-    if (assets?.length > 0) onLink(assets[0]);
-    hideUploadModal();
-  }, [createAssets, fileList, hideUploadModal, onLink, setUploading]);
+    let assets: (Asset | undefined)[] = [];
+    let asset: Asset | undefined;
```
KaWaite, line 34: I can see what you are doing, but this callback is `handleUpload`. It shouldn't also have logic around linking. Using `asset` as the way to know whether to link or not isn't obvious. 

Also, is the only reason you are saving `uploadUrl` to state to be used in this callback? If so, please just pass the values to the callback directly. If you have issues around getting the value from antd's form or soemthing, and that is the main reason you decided to use state, just make the state locally in the form's component and pass the values of that state to this callback as props.

web/src/components/molecules/Asset/AssetList/index.tsx
```diff
@@ -56,18 +70,32 @@ const AssetList: React.FC<Props> = ({
...
-    setUploadModalVisibility(false);
-    setUploading(false);
-    setFileList([]);
-  }, [setFileList, setUploading, setUploadModalVisibility]);
-
-  const handleUpload = useCallback(() => {
+  const handleUpload = useCallback(async () => {
     setUploading(true);
-    onAssetCreate(fileList).finally(() => {
+
```
KaWaite, line 75: On line 74, you setUploading to true, but then it seems like you don't set it to false when finished

KaWaite: Also, is uploading/setUploading only used here? We have a lot of props being passed as it stands, and it would be nice to keep state exactly where it is needed.
If it is only needed here, in AssetList/index.tsx, consider moving the declaration here (and any other state/props that aren't needed so high up, and then creating a local handleOnModalClose or whatever callback is needed to setUploading(false)

web/src/components/molecules/Schema/FieldModal/FieldCreationModal/index.tsx
nourbalaha: Done!

web/src/components/molecules/Asset/AssetList/index.tsx
nourbalaha: Done!

web/src/components/organisms/Asset/AssetList/hooks.ts
nourbalaha: Done!

web/src/components/organisms/Asset/AssetList/hooks.ts
nourbalaha: Done!

web/src/components/molecules/Common/Form/AssetItem/hooks.ts
nourbalaha: The error has been handles from these two functions `onAssetsCreate` and `onAssetCreateFromUrl`.
in this function I just want to hide the upload modal if an error happened.

if I put the `hideUploadModal` function outside the switch statement it will not run if an error happened.

web/src/components/molecules/Asset/AssetList/index.tsx
nourbalaha: setUploading to false is done inside `hideUploadModal` function.

`uploading` and `setUploading` are used in multiple places.

nourbalaha: Anyways, I've just made a refactor to use setUploading only from one place.

web/src/components/molecules/Common/Form/AssetItem/hooks.ts
nourbalaha: Please let me know what do you think.

web/src/components/molecules/Common/Form/AssetItem/hooks.ts
nourbalaha: Please let me know what do you think.

web/src/components/organisms/Asset/AssetList/hooks.ts
KaWaite: I see you updated it to handleUploadModalCancel. I get that closing an update modal is kinda like "cancelling" the update, but in actual fact the callback is handling the close of the modal, right? So can you make it "close"

KaWaite: I think you have updated other callbacks too with xxxCancel. Can you do the same for those too?

web/src/components/molecules/Common/Form/AssetItem/hooks.ts
KaWaite: Just a couple things.
If the error is being handled in those functions, I don't think you need the try/catch, since you aren't catching anything.
Also, probably you just mistakenly wrote switch instead of try/catch above, but hideUploadModal is inside the try/catch, so by removing the try/catch you kinda solve the issue anyways :P

web/src/components/molecules/Asset/AssetList/index.tsx
```diff
@@ -42,32 +45,32 @@ const AssetList: React.FC<Props> = ({
...
-      hideUploadModal();
-    });
-  }, [fileList, setUploading, onAssetCreate, hideUploadModal]);
+  const handleUpload = useCallback(async () => {
+    if (uploadType === "url") {
+      await onAssetCreateFromUrl(uploadUrl);
+    } else {
+      await onAssetsCreate(fileList);
+    }
+  }, [uploadType, onAssetCreateFromUrl, uploadUrl, onAssetsCreate, fileList]);
```
KaWaite, line 73: If you are using `onAssetCreateFromUrl` and `onAssetsCreate` only in this callback, can you move the callback up to where the `uploadUrl` is declared and then pass just `onUpload={handleUpload}`?

KaWaite: Also, does this work correctly? I see you are using handleUpload like `const asset = handleUpload()` Do you need to do something like this to have something returned?
```
const result = await onAssetCreateFromUrl(uploadUrl);
return result
```

nourbalaha: For `AssetList` page I use this function and I do not need to return anything since there is no linking involved from there.
https://github.com/reearth/reearth-cms/blob/a1c09d2f850f0b257e98aa7f219d6573dce8fea1/web/src/components/molecules/Asset/AssetList/index.tsx#L67

However, for `LinkAssetModal` that is used in `Schema` and `Content` pages I use a different function called `handleUploadAndLink`
https://github.com/reearth/reearth-cms/blob/a1c09d2f850f0b257e98aa7f219d6573dce8fea1/web/src/components/molecules/Common/Form/AssetItem/hooks.ts#L38

Both of these function are passed to the same `UploadAsset` component.

PR #439, opened 2022-11-21
fix(server): text length is incorrect for Japanese characters 

mimoham24:
# Overview
When creating/updating an item, long descriptions in Japanese give an error
## What I've done
check by `[]rune{})`


rot1024:
Fixed by 2a7456fde45f2c2ace06d86b78b76fa5b5ad61ab

PR #440, opened 2022-11-22
refactor(server): remove workspace from the item query

mimoham24:


server/pkg/item/query.go
```diff
@@ -3,16 +3,14 @@ package item
 import "github.com/reearth/reearth-cms/server/pkg/id"
 
 type Query struct {
-	workspace id.WorkspaceID
-	project   id.ProjectID
-	q         string
+	project id.ProjectID
```
rot1024, line 6: Did you update `schemas/integration.yml`?

rot1024 (changes requested):

server/pkg/item/query.go
mimoham24: I don't believe that we have a query integration yet

rot1024 (approved):

PR #443, opened 2022-11-22
fix(server): disable sanitizing file names on GCS

nourbalaha:
# Overview
I this PR I fixed GCS's object names when uploading an asset by removing the function that replaces special characters with dashes.
This function is no longer needed since `sanitize.Path()` method was removed from `ReadAsset()` method.


rot1024 (commented):
~~Can you disable reaming in `inftrastructure/fs` also?~~

rot1024 (approved):

PR #446, opened 2022-11-24
feat(web): add unique indicator

issmail-basel:
# Overview
Add a field indicator on both the "Schema" screen and "content form" screen, as displayed in Figma.

## Screenshots
Items form:

Schema screen:


web/src/components/molecules/Content/Form/index.tsx
```diff
@@ -123,7 +123,12 @@ const ContentForm: React.FC<Props> = ({
                 },
               ]}
               name={field.id}
-              label={field.title}>
+              label={
+                <>
+                  <FieldTitle>{field.title}</FieldTitle>
+                  {field.unique ? <FieldUnique>({t("unique")})</FieldUnique> : ""}
+                </>
```
KaWaite, line 130: To tidy up this component a bit, can you create a small component (can be inside the same file) for `FieldTitle` that takes in a title and isUnique prop and returns this fragment? It is a lot of duplicated code that can be reduced to 
`<FieldTitle title={field.title} isUnique={field.unique} />`

web/src/components/molecules/Schema/ModelFieldList.tsx
```diff
@@ -121,4 +125,10 @@ const ItemKey = styled.span`
   font-weight: 400;
 `;
 
+const ItemUnique = styled.span`
```
KaWaite, line 128: Is there a reason you are using spans for all the text and not, for example, a p tag?

KaWaite: It is just that span usually isn't used as the main tag for text. Usually it is used for styling sections of text, which is why I am not sure its a good idea to use it as a text element.
Of course using something like p you have to remove the margin, but since it is more meaningful syntactically I think it is better to use p or another text element and then change the size if necessary.
To be honest, unless it is a header or subheader (where h1 or h2 are obvious), I typically just use p and remove the margin. Not great, but still relatively syntactically meaningful

issmail-basel: Now the name is a `p` tag, and the others (unique indicator, etc..) are children of `p` tag so I kept them as `span`

KaWaite (approved):

PR #455, opened 2022-11-29
test(server): add test to integration asset

gop50k:
# Overview


server/e2e/integration_asset_test.go
```diff
@@ -0,0 +1,71 @@
...
+		Expect().
+		Status(http.StatusNotFound)
+
+	e.GET("/api/assets/{assetId}", aid).
+		WithHeader("authorization", "Bearer "+secret).
+		Expect().
+		Status(http.StatusOK).
+		JSON().
+		Object().Keys().
+		Contains("id", "projectId", "name", "url", "contentType", "previewType", "totalSize", "file", "createdAt")
```
rot1024, line 35: Checking only for the presence or absence of a key does not allow you to immediately notice when a problem occurs with the contents. The easiest way is to compare the whole object with the Equal method, but that may be difficult. In that case, perform an assertion only on comparable objects as follows.

```go
	obj = e.GET("/api/models/{modelId}/items", ikey).
		WithHeader("authorization", "Bearer "+secret).
		WithQuery("page", 1).
		WithQuery("perPage", 5).
		Expect().
		Status(http.StatusOK).
		JSON().
		Object()

	obj.Value("page").Equal(1)
	obj.Value("perPage").Equal(5)
	obj.Value("totalCount").Equal(1)

	a = obj.Value("items").Array()
	a.Length().Equal(1)
	a.First().Object().Value("id").Equal(itmId.String())
	a.First().Object().Value("fields").Equal([]any{})
	a.First().Object().Value("parents").Equal([]any{})
	a.First().Object().Value("refs").Equal([]string{"latest"})
```

rot1024 (changes requested):

rot1024 (approved):

PR #463, opened 2022-12-01
feat(web): delete my integrations

issmail-basel:
# Overview
Added deletion functionality to the Integrations on the "my integration" screen.

## What I've done
- Add danger zone to my integration form
- Add deletion queries/hooks


web/src/components/molecules/Integration/types.ts
```diff
@@ -6,7 +6,7 @@ export type IntegrationMember = {
   active: boolean;
 };
 
-export type Role = "WRITER" | "READER" | "OWNER";
+export type Role = "WRITER" | "READER" | "OWNER" | "Maintainer";
```
KaWaite, line 9: I think I saw earlier you discussing this, but is it correct that it is `Maintainer` and not `MAINTAINER` because of a mistake on the backend? If so, please resolve this message

web/src/i18n/translations/en.yml
```diff
@@ -111,8 +111,14 @@ New Integration: ''
 Create: ''
 Integration Name: ''
 Please input the title of the integration!: ''
-Integration Token: ''
 Create new integration: ''
+Are you sure remove this intergration?: ''
+Permanently remove your Intergration and all of its contents from the Re:Earth CMS.: ''
```
KaWaite, line 116:
```suggestion
Permanently remove your Integration and all of its contents from the Re:Earth CMS.: ''
```

web/src/i18n/translations/en.yml
```diff
@@ -111,8 +111,14 @@ New Integration: ''
...
+Once intergration was removed, it will disappear from all workspace.: ''
+Remove Intergration: ''
+Permanently remove your Intergration and all of its contents from the Re:Earth CMS. This action is not reversible – please continue with caution.: ''
```
KaWaite, line 119:
```suggestion
Once the integration is removed, it will disappear from all workspaces.: ''
Remove Integration: ''
Permanently remove your Integration and all of its contents from the Re:Earth CMS. This action is not reversible – please continue with caution.: ''
```

KaWaite (approved):

web/src/components/molecules/Integration/types.ts
issmail-basel: Yes this is because of the backend.

PR #465, opened 2022-12-02
fix(web): asset upload modal not closing after errors

nourbalaha:
# Overview
In this PR I fixed asset upload modal not closing after errors bug.

### DIFF
--- a/web/src/components/organisms/Asset/AssetList/hooks.ts
+++ b/web/src/components/organisms/Asset/AssetList/hooks.ts
@@ -96,26 +96,28 @@ export default () => {
       (async () => {
         if (!projectId) return [];
         setUploading(true);
-        const results = (
-          await Promise.all(
-            files.map(async file => {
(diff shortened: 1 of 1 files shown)


web/src/components/organisms/Asset/AssetList/hooks.ts
```diff
@@ -96,26 +96,28 @@ export default () => {
...
+                return convertAsset(result.data.createAsset.asset as GQLAsset);
+              }),
+            )
+          ).filter(Boolean);
+          if (results?.length > 0) {
+            Notification.success({ message: t("Successfully added one or more assets!") });
+            await refetch();
+          }
+        } catch {
+          Notification.error({ message: t("Failed to add one or more assets.") });
```
KaWaite, line 117: Like in our discussions with your other PR, does this error ever show? Since you the Promise.all is being resolved to `results`, doesn't that mean we will never hit this catch, making the try/catch not needed.

PR #469, opened 2022-12-04
feat(server): support multiple value

yk-eukarya:


server/pkg/schema/field.go
```diff
@@ -146,12 +148,22 @@ func (f *Field) Clone() *Field {
...
+func (f *Field) Validate(vs *value.Multiple) error {
+	if f.required && (vs.IsEmpty() || vs.First().IsEmpty()) {
 		return ErrValueRequired
 	}
-	if v.IsEmpty() {
+	if vs.IsEmpty() {
 		return nil
 	}
-	return f.typeProperty.Validate(v)
+	if !f.multiple && len(vs.Values()) > 1 {
```
rot1024, line 160:
```suggestion
	if !f.multiple && vs.Len() > 1 {
```

Impl `func (*value.Multiple) Len() int` method to prevent cloning a slice of values.

server/internal/adapter/gql/loader_item.go
```diff
@@ -32,7 +33,7 @@ func (c *ItemLoader) Fetch(ctx context.Context, ids []gqlmodel.ID) ([]*gqlmodel.
 	}
 
 	return lo.Map(res, func(m item.Versioned, i int) *gqlmodel.Item {
-		return gqlmodel.ToItem(m.Value())
+		return gqlmodel.ToItem(m.Value(), false)
```
rot1024, line 36:
```suggestion
		return gqlmodel.ToItem(m.Value(), sf.Multiple())
```

1. Get `[]id.SchemaID` from the result of `c.usecase.FindByIDs` (item.List)
2. Load schemas from the ID slice.
3. Set the schema field is `multiple` or not to the second argument of `ToItem` func with the corresponding schema field.

server/internal/adapter/gql/resolver_mutation_item.go
```diff
@@ -46,7 +46,7 @@ func (r *mutationResolver) UpdateItem(ctx context.Context, input gqlmodel.Update
 	}
 
 	return &gqlmodel.ItemPayload{
-		Item: gqlmodel.ToItem(res.Value()),
+		Item: gqlmodel.ToItem(res.Value(), false),
```
rot1024, line 49:
```suggestion
		Item: gqlmodel.ToItem(res.Value(), sf.Multiple()),
```

rot1024 (changes requested):
The general policy is OK. Just don't forget to add a unit test for the newly added method or function that will be a multioh!

server/pkg/integrationapi/item.go
```diff
@@ -28,12 +29,13 @@ func NewVersionedItem(ver item.Versioned) VersionedItem {
 	}
 }
 
-func NewItem(i *item.Item) Item {
+func NewItem(i *item.Item, s *schema.Schema) Item {
 	fs := lo.Map(i.Fields(), func(f *item.Field, _ int) Field {
+		sf := s.Field(f.FieldID())
```
rot1024, line 34: Note that `sf` can be nil. If nil, drop the current field (use `lo.FilterMap` instead of `lo.Map`)

rot1024 (approved):
I would like to see more unit tests. Please write unit tests with the intention of achieving 100% code coverage, especially for the domain model. Get into the habit of writing implementation and unit tests at the same time.

PR #471, opened 2022-12-06
feat(server): integration API can accept keys as model IDs

rot1024:
- API `GET,POST /models/{modelId}/items` now possible to retrieve items by specifying the model key instead of the model ID.
- All APIs now possible to return 404 error.
- Add `key` field to item fields


server/internal/infrastructure/mongo/model.go
```diff
@@ -84,6 +85,24 @@ func (r *Model) FindByKey(ctx context.Context, projectID id.ProjectID, key strin
...
+	if key != nil {
+		filters = append(filters, bson.M{"key": key.String()})
+	}
+	if len(filters) == 0 {
+		return nil, rerror.ErrNotFound
+	}
+
+	return r.findOne(ctx, bson.M{
+		"project": bson.M{"$in": pids.Strings()},
+		"$or":     filters,
```
yk-eukarya, line 102: Maybe we need to use `$and` to insure the premissions

HideBa (approved):
I think it is fine in the source code. On the other hand, I have one concern.

What if a malicious user sets someone else's Model ID as a key? I think such a situation is unlikely to happen, but if a user does so, he/she may be able to retrieve another person's Item.
To prevent this, I think it is necessary to validate the string that can be successfully parsed as a ULID when creating a key, or to check if the ID is not set as the ID of another Model.

server/internal/infrastructure/mongo/model.go
rot1024: Actually it works. See e2e test

rot1024:
> I think it is fine in the source code. On the other hand, I have one concern.
>
> What if a malicious user sets someone else's Model ID as a key? I think such a situation is unlikely to happen, but if a user does so, he/she may be able to retrieve another person's Item.
> To prevent this, I think it is necessary to validate the string that can be successfully parsed as a ULID when creating a key, or to check if the ID is not set as the ID of another Model.

Fixed. If the key can be parsable as an ID, it never be treated as a key.

PR #490, opened 2022-12-11
feat(web): delete items

issmail-basel:
# Overview
Add the `delete` option to the content table so you can `delete` one or more items from the list of added items

## What I've done
- Hook to remove the item from the workspace
- Options to select/deselect items from the table

## What I haven't done
No changes in the project structure

## How I tested
- Create a new model and new schema
- Go to the items screen
- Add one or more item
- Select/Deselect items from the table
- Click on Delete (items should be removed)


web/src/components/molecules/Content/List/index.tsx
```diff
@@ -20,11 +21,16 @@ export type Props = {
   contentTableColumns?: ProColumns<ContentTableField>[];
   modelsMenu: React.ReactNode;
   selectedItem: Item | undefined;
+  selection: {
+    selectedRowKeys: Key[];
```
KaWaite, line 25: Is selectedRowKeys sometimes a string AND sometimes a number? Usually in this situation we either use, for example field IDs, which are strings, or the index while we map through another array, which is always a number.
If possible, just use a simple type instead of the Key type from React.

KaWaite (approved):

web/src/components/molecules/Content/Table/index.tsx
```diff
@@ -61,6 +74,29 @@ const ContentTable: React.FC<Props> = ({
     [onItemEdit, onItemSelect, selectedItem?.id],
   );
 
+  const rowSelection: TableRowSelection = {
+    selectedRowKeys: selection.selectedRowKeys,
+    onChange: (selectedRowKeys: any) => {
```
nourbalaha, line 79: Can you please add a type here?

web/src/components/molecules/Content/Table/index.tsx
```diff
@@ -61,6 +74,29 @@ const ContentTable: React.FC<Props> = ({
...
+      setSelection({
+        ...selection,
+        selectedRowKeys: selectedRowKeys,
+      });
+    },
+  };
+
+  const AlertOptions = (props: any) => {
```
nourbalaha, line 87: Can you please add a type here?

nourbalaha (approved):

web/src/components/molecules/Content/List/index.tsx
issmail-basel: @KaWaite , it is used like this in multiple places now, we are using `Key` following and docs, but I think it can be removed.

PR #491, opened 2022-12-11
feat(web): disable the comment button when no item is selected

issmail-basel:
# Overview
- disable the comment button when no item is selected
- Stop comment action spinner for readers


KaWaite (approved):
Please make the fix before merging

PR #502, opened 2022-12-12
feat(server): update version validation

mimoham24:
# Overview
- Item can belong to multiple request
- When request is approved, the latest version will be published
- If approved version is already on public, no change and no error
- When request is created for an item whose latest version is public, it should returns error
## What I've done
- When request is approved, the latest version will be published
- If approved version is already on public, no change and no error
## What I haven't done
already implemented:
- Item can belong to multiple request
- When request is created for an item whose latest version is public, it should returns error

### DIFF
--- a/server/internal/usecase/interactor/request.go
+++ b/server/internal/usecase/interactor/request.go
@@ -219,8 +219,12 @@ func (r Request) Approve(ctx context.Context, requestID id.RequestID, operator *
 
 	// apply changes to items (publish items)
 	for _, item := range req.Items() {
-		// publish the approved version
-		if err := r.repos.Item.UpdateRef(ctx, item.Item(), version.Public, item.Pointer().Ref()); err != nil {
+		// If approved version is already on public, no change and no error
+		if item.Pointer().IsRef(version.Public) {
+			continue
+		}
+		// publish the latest version
+		if err := r.repos.Item.UpdateRef(ctx, item.Item(), version.Public, version.Latest.OrVersion().Ref()); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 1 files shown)


server/internal/usecase/interactor/request.go
```diff
@@ -219,8 +219,12 @@ func (r Request) Approve(ctx context.Context, requestID id.RequestID, operator *
 
 	// apply changes to items (publish items)
 	for _, item := range req.Items() {
-		// publish the approved version
-		if err := r.repos.Item.UpdateRef(ctx, item.Item(), version.Public, item.Pointer().Ref()); err != nil {
+		// If approved version is already on public, no change and no error
+		if item.Pointer().IsRef(version.Public) {
+			continue
+		}
```
rot1024, line 225:
```suggestion
```

This implementation has no meaningful behavior; the pointer in items of a Request is always "latest".

rot1024 (changes requested):

server/internal/usecase/interactor/request.go
```diff
@@ -219,10 +219,6 @@ func (r Request) Approve(ctx context.Context, requestID id.RequestID, operator *
 
 	// apply changes to items (publish items)
 	for _, item := range req.Items() {
-		// If approved version is already on public, no change and no error
-		if item.Pointer().IsRef(version.Public) {
-			continue
-		}
 		// publish the latest version
 		if err := r.repos.Item.UpdateRef(ctx, item.Item(), version.Public, version.Latest.OrVersion().Ref()); err != nil {
```
rot1024, line 223:
```suggestion
		if err := r.repos.Item.UpdateRef(ctx, item.Item(), version.Public, item.Pointer().Ref()); err != nil {
```

rot1024 (approved):

PR #513, opened 2022-12-15
fix(web): integer field min, max and default value

nourbalaha:
# Overview
In this PR I fixed the following:
- When integer field value is being created or updated with default values existed the frontend sends null.
- When min, max and default values are null the frontend sends 0 instead of null.

### DIFF
--- a/web/src/components/molecules/Common/MultiValueField/index.tsx
+++ b/web/src/components/molecules/Common/MultiValueField/index.tsx
@@ -29,7 +29,7 @@ const MultiValueField: React.FC<Props> = ({
     (e: ChangeEvent<HTMLInputElement>, id: number) => {
       onChange?.(
         value?.map((valueItem, index) =>
-          index === id ? (typeof e === "number" ? e : e.target.value) : valueItem,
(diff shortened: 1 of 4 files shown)


web/src/components/molecules/Common/MultiValueField/index.tsx
```diff
@@ -29,7 +29,7 @@ const MultiValueField: React.FC<Props> = ({
     (e: ChangeEvent<HTMLInputElement>, id: number) => {
       onChange?.(
         value?.map((valueItem, index) =>
-          index === id ? (typeof e === "number" ? e : e.target.value) : valueItem,
+          index === id ? (typeof e === "number" ? e : e?.target.value) : valueItem,
```
KaWaite, line 32: If e can be undefined, please update the prop type too to keep types consistent (right now it is `e: ChangeEvent<HTMLInputElement>`, but should be `e?: Change.....`)

KaWaite (approved):

web/src/components/molecules/Common/MultiValueField/index.tsx
nourbalaha: Done!

PR #515, opened 2022-12-16
feat(server): update date fields of requests, GraphQL request resolvers

mimoham24:
# Overview
- update `UpdateAt` on request update
- update `ApprovedAt` and `ClosedAt` on state change
- fix reviewers resolver
- add item resolver to RequestItem


server/internal/adapter/gql/resolver_request.go
```diff
@@ -25,13 +25,35 @@ func (r requestResolver) Project(ctx context.Context, obj *gqlmodel.Request) (*g
 }
 
 func (r requestResolver) Reviewers(ctx context.Context, obj *gqlmodel.Request) ([]*gqlmodel.User, error) {
-	res, err := dataloaders(ctx).User.LoadAll(obj.ReviewersID)
-	if len(err) > 0 {
-		return nil, err[0]
-	}
-	return res, nil
+	res, errors := dataloaders(ctx).User.LoadAll(obj.ReviewersID)
+	return res, errors[0]
```
rot1024, line 29: Unnecessary change: if errors is nil or empty, it panics

server/schemas/request.graphql
```diff
@@ -25,6 +25,8 @@ type RequestItem {
   version: String
   ref: String
   item: VersionedItem
+  schema: Schema
+  model: Model
```
rot1024, line 29:
```suggestion
```

Add these fields and resolvers to the Item type instead of adding it here.

rot1024 (changes requested):

server/schemas/request.graphql
mimoham24: Do you mean the versioned Item?

server/internal/adapter/gql/resolver_request.go
mimoham24: `errors` always [nil]

server/schemas/request.graphql
rot1024: No, `Item` type

server/internal/adapter/gql/resolver_request.go
rot1024: Future implementation changes may cause an error to be returned. There is no particular harm in handling errors, so they should be left as they are.

rot1024 (approved):

rot1024 (approved):

PR #526, opened 2022-12-20
feat(web): add tooltip to header menu items

issmail-basel:
# Overview
Add a tooltip when hovering over the workspace name on the header menu items

### DIFF
--- a/web/src/components/molecules/Common/Header/index.tsx
+++ b/web/src/components/molecules/Common/Header/index.tsx
@@ -6,6 +6,7 @@ import { useAuth } from "@reearth-cms/auth";
 import Header from "@reearth-cms/components/atoms/Header";
 import Icon from "@reearth-cms/components/atoms/Icon";
 import Menu from "@reearth-cms/components/atoms/Menu";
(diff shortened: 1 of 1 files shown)


nourbalaha (approved):
Approved!

web/src/components/molecules/Common/Header/index.tsx
```diff
@@ -65,7 +66,11 @@ const HeaderMolecule: React.FC<Props> = ({
           children: workspaces
             ?.filter(workspace => workspace.id === personalWorkspace?.id)
             ?.map(workspace => ({
-              label: <MenuText>{workspace.name}</MenuText>,
+              label: (
+                <Tooltip title={workspace.name} placement="right">
+                  <MenuText>{workspace.name}</MenuText>
```
KaWaite, line 71: Is there any point making a tooltip that just shows what the menu text already shows? Typically a tooltip would have a description or something different than the original, no?

issmail-basel: Yup, because the workspace name can be very long (it will end with three dots), so I've been asked to have a tooltip for the name.

KaWaite (approved):

PR #529, opened 2022-12-21
fix(web): tileset preview + refactor

nourbalaha:
# Overview
In this PR I did the following:
1- Fixed tileset preview full screen button for `GEO3D` and `MODEL3D` types.
2- Hide full screen button with `Unsupported` viewer type.
3- I refactored the way the asset viewer is displayed.

### DIFF
--- a/web/src/components/molecules/Asset/Asset/AssetBody/Asset.tsx
+++ b/web/src/components/molecules/Asset/Asset/AssetBody/Asset.tsx
@@ -6,7 +6,7 @@ import DownloadButton from "@reearth-cms/components/atoms/DownloadButton";
 import { DefaultOptionType } from "@reearth-cms/components/atoms/Select";
 import TilesetPreview from "@reearth-cms/components/atoms/TilesetPreview";
 import UserAvatar from "@reearth-cms/components/atoms/UserAvatar";
-import { Asset } from "@reearth-cms/components/molecules/Asset/asset.type";
+import { Asset, ViewerType } from "@reearth-cms/components/molecules/Asset/asset.type";
 import Card from "@reearth-cms/components/molecules/Asset/Asset/AssetBody/card";
 import PreviewToolbar from "@reearth-cms/components/molecules/Asset/Asset/AssetBody/previewToolbar";
 import {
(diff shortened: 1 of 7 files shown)


web/src/components/organisms/Asset/Asset/hooks.ts
```diff
@@ -3,16 +3,24 @@ import { useCallback, useEffect, useMemo, useState } from "react";
...
   PreviewType as GQLPreviewType,
   useGetAssetQuery,
   useUpdateAssetMutation,
 } from "@reearth-cms/gql/graphql-client-api";
 import { useT } from "@reearth-cms/i18n";
+import { getExtension } from "@reearth-cms/utils/file";
 
 import { convertAsset } from "../convertAsset";
 
+export type ViewerType = "TileSet" | "Image" | "SVG" | "Unsupported";
```
KaWaite, line 22: I think these are used code-side only, right? Since they aren't shown to the user or matched with the same values from backend values, etc, you don't need to capitalize. Keeping simple makes typos less likely.
```
export type ViewerType = "tileset" | "image" | "svg" | "unsupported";
```

KaWaite: Also, I don't think unsupported is needed. It is just a fallback

web/src/components/molecules/Asset/Asset/AssetBody/Asset.tsx
```diff
@@ -17,13 +17,8 @@ import SideBarCard from "@reearth-cms/components/molecules/Asset/Asset/AssetBody
 import UnzipFileList from "@reearth-cms/components/molecules/Asset/Asset/AssetBody/UnzipFileList";
 import ViewerNotSupported from "@reearth-cms/components/molecules/Asset/Asset/AssetBody/viewerNotSupported";
 import ArchiveExtractionStatus from "@reearth-cms/components/molecules/Asset/AssetListTable/ArchiveExtractionStatus";
-import {
-  fileFormats,
-  imageFormats,
-  compressedFileFormats,
-} from "@reearth-cms/components/molecules/Common/Asset";
+import { ViewerType } from "@reearth-cms/components/organisms/Asset/Asset/hooks";
```
KaWaite, line 20: Shouldn't have cyclical imports. You should define this type in the molecule, and then import it, with the component, to the organism

web/src/components/organisms/Asset/Asset/hooks.ts
nourbalaha: Done!

Now that I am thinking about it I renamed "tileset" to "cesium" since it is more general and the viewer supports 3dtilesets and other formats like "czml" and "kml"... 

Also I think that I will keep "unsupported" since it is used here:
https://github.com/reearth/reearth-cms/blob/999e8dbe034d1f7350fbabb68ad1ed199277c6b0/web/src/components/molecules/Asset/Asset/AssetBody/previewToolbar.tsx#L28

Please let me know what do you think.

web/src/components/molecules/Asset/Asset/AssetBody/Asset.tsx
nourbalaha: Done!

KaWaite (approved):

PR #531, opened 2022-12-22
feat(web): show caution message when user deletes schema field

issmail-basel:
# Overview
Show caution message when the user deletes schema field


web/src/components/molecules/Schema/ModelFieldList.tsx
```diff
@@ -32,7 +48,7 @@ const ModelFieldList: React.FC<Props> = ({
           actions={[
             <Icon
               icon="delete"
-              onClick={() => handleFieldDelete((item as Field).id)}
+              onClick={() => handleFieldDeleteConfirmation((item as Field).id)}
```
nourbalaha, line 51: Just a suggestion, you can change line 102 to `const FieldStyledList = styled(List<Field>)` and then:
```suggestion
              onClick={() => handleFieldDeleteConfirmation(item.id)}
```

nourbalaha (approved):

web/src/components/molecules/Schema/ModelFieldList.tsx
issmail-basel: @nourbalaha , there is an issue with `antd` it can't predict the type I'm passing to `renderItem`

KaWaite (approved):

PR #534, opened 2022-12-22
feat(server): item sorting 

mimoham24:
# Overview
- item sorting
- add schema to the item query


server/pkg/sort/direction.go
```diff
@@ -0,0 +1,28 @@
+package sort
```
rot1024, line 1: too abstract packaga name. also it conflicts with `sort` package in golang. DIrection type shold be implemented in `item` package because it is currently only used in `item` package.

rot1024 (changes requested):

rot1024 (approved):

PR #537, opened 2022-12-22
feat(server): customized order of schema fields

mimoham24:
# Overview
user can now specify the schema fields' order 
## What I've done
- added a new attribute (order)
- auto increment of order when adding a new field
- order fields according to the order number


server/schemas/field.graphql
```diff
@@ -21,6 +21,7 @@ type SchemaField {
   typeProperty: SchemaFieldTypeProperty
   key: String!
   title: String!
+  order: Int!
```
rot1024, line 24:
```suggestion
  order: Int
```

If there is no order, it should be automatically decided

rot1024 (changes requested):

server/schemas/field.graphql
mimoham24: it's already being automatically decided. thank you

rot1024 (approved):

PR #539, opened 2022-12-24
feat(web): redesign accessibility screen

issmail-basel:
# Overview
Fix the design of the accessibility screen


web/src/components/molecules/Accessibility/index.tsx
```diff
@@ -151,59 +154,46 @@ const Public: React.FC<Props> = ({
 
   const publicScopeList = [
     { id: 1, name: t("Private"), value: "private" },
-    // {
-    //   id: 2,
-    //   name: t("API Key Only"),
-    //   value: "limited",
-    // },
     { id: 3, name: t("Public"), value: "public" },
```
nourbalaha, line 157:
```suggestion
    { id: 2, name: t("Public"), value: "public" },
```

nourbalaha (approved):

PR #541, opened 2023-01-06
feat(web): show asset name instead of id

issmail-basel:
# Overview
Show asset name instead of id inside the request details screen


issmail-basel:
Request UI changed

PR #551, opened 2023-01-11
feat(server): offset based pagination support

yk-eukarya:


server/internal/adapter/gql/gqlmodel/convert.go
```diff
@@ -25,6 +25,12 @@ func (p *Pagination) Into() *usecasex.Pagination {
 	if p == nil {
 		return nil
 	}
+	if p.Offset != nil {
+		return usecasex.OffsetPagination{
+			Offset: int64(*p.Offset),
+			Limit:  int64(*p.First),
```
rot1024, line 31: If fiest is nil, it panics. If it's nil, set default limit (50)

rot1024 (changes requested):

rot1024 (approved):

PR #570, opened 2023-01-14
feat(server): update multiple fields

mimoham24:
# Overview

## What I've done
add `updateFields `mutation to update many fields at once


server/internal/adapter/gql/resolver_mutation_field.go
```diff
@@ -125,3 +128,46 @@ func (r *mutationResolver) DeleteField(ctx context.Context, input gqlmodel.Delet
...
+	ms, err := usecases(ctx).Model.FindByIDs(ctx, []id.ModelID{mId}, getOperator(ctx))
+	if err != nil || len(ms) != 1 || ms[0].ID() != mId {
+		if err == nil {
+			return nil, errors.New("not found")
+		}
+		return nil, err
+	}
+	m := ms[0]
+
+	s, err := usecases(ctx).Schema.FindByID(ctx, m.Schema(), getOperator(ctx))
```
yk-eukarya, line 158: IMO there is no need to load the schema here, you can just pass the `m.Schema()` to `UpdateFieldsOrder`

server/internal/usecase/interactor/schema.go
```diff
@@ -88,53 +88,10 @@ func (i Schema) UpdateField(ctx context.Context, param interfaces.UpdateFieldPar
...
-		}
-
-		if param.Unique != nil {
-			f.SetUnique(*param.Unique)
-		}
-
-		if param.Multiple != nil {
-			f.SetMultiple(*param.Multiple)
+		err = updateField(param, f)
+		if err != nil {
```
rot1024, line 92:
```suggestion
		if err := updateField(param, f); err != nil {
```

rot1024 (approved):

PR #572, opened 2023-01-15
perf: decompression rate improvement

pyshx:
This commit provides significant improvements to the existing decompression time but it has some caveats including more RAM usage, abandonment to the clean architecture etc. Keeping in mind the PubSub timeout limit it is estimated to have upper limit of 4-5GBs of decompression capacity at the moment hence its a temporary solution.

# Overview
Decompression rate improvement using concurrency

## What I've done
I've implement a decompression rate improvement solution of BigFiles stored at GCS

## How I tested
I tested it by firing up the worker locally and making a decompression request at the `/api/decompress` endpoint via postman.

### DIFF
--- a/worker/pkg/decompressor/decompressor.go
+++ b/worker/pkg/decompressor/decompressor.go
@@ -6,12 +6,15 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"path/filepath"
 	"strings"
 	"sync"
 
 	"cloud.google.com/go/storage"
 	"github.com/bodgit/sevenzip"
+	"github.com/joho/godotenv"
+	"github.com/kelseyhightower/envconfig"
 	"github.com/reearth/reearthx/log"
 )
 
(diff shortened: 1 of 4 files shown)


worker/pkg/decompressor/decompressor.go
```diff
@@ -15,6 +21,12 @@ var (
 
 const limit = 1024 * 1024 * 1024 * 30 // 30GB
 
+const (
+	GCS_BUCKET_NAME               = "asset.cms.test.reearth.dev"
```
rot1024, line 25: Do not hard-code

rot1024 (changes requested):

worker/pkg/decompressor/decompressor.go
pyshx: Ok, I wanted to keep the changes to minimum as this is a temporary solution.
Is it okay if i use the `.env` variables instead?

PR #590, opened 2023-01-17
fix(web): viewer re-rendering + refactor

nourbalaha:
# Overview
In this PR I fixed the bug that causes the viewer to re-render on every click and I did some refactoring


HideBa (approved):
LGTM

PR #598, opened 2023-01-19
style(web): basic styling for infobox

nourbalaha:
# Overview
I did a basic styling for InfoBox based on the figma designs

### DIFF
--- a/web/src/components/molecules/Asset/InfoBox/index.tsx
+++ b/web/src/components/molecules/Asset/InfoBox/index.tsx
@@ -62,6 +62,7 @@ const InfoBoxWrapper = styled.div`
 `;
 
 const Header = styled.div`
+  height: 50px;
   display: flex;
   justify-content: space-between;
   align-items: center;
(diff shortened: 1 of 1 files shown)


web/src/components/molecules/Asset/InfoBox/index.tsx
```diff
@@ -58,25 +57,34 @@ const InfoBoxWrapper = styled.div`
...
 const Header = styled.div`
+  height: 50px;
   display: flex;
   justify-content: space-between;
   align-items: center;
+  padding: 8px 12px;
+  box-shadow: inset 0px -1px 0px #f0f0f0;
 `;
 
 const Title = styled.div`
```
issmail-basel, line 73: @nourbalaha , I think you can remove the property from the:
` <Title color={theme.base01}>{title}</Title>`
color is no longer used here.

issmail-basel (approved):

web/src/components/molecules/Asset/InfoBox/index.tsx
nourbalaha: Makes sense, thanks!

PR #602, opened 2023-01-20
fix(worker): remove colliding dirname duplication from decompressed zipfile & refactor

pyshx:
# Overview
This small PR fixes an issue with the GCS uploaded filepath and remove numWorkers & workerQueueSize from `.env`.

Take a look at this [func](https://github.com/reearth/reearth-cms/blob/9a51343bb82f753e6a3221f305af256128cae57e/worker/pkg/decompressor/decompressor.go#L168) in decompressor.go


HideBa (approved):
Look good for me! Thanks. 
Later I'll refactor these code base so that it won't depend on GCS.

PR #611, opened 2023-01-25
feat(server): skip asset decompression option

yk-eukarya:
# Overview
I've implemented the skip item decompression for creating asset
- in create asset gql mutation
- in create asset integration api
- some enhancement for integration openAPI schema


server/internal/usecase/interactor/asset.go
```diff
@@ -137,32 +143,41 @@ func (i *Asset) Create(ctx context.Context, inp interfaces.CreateAssetParam, op
...
+	if err := i.gateways.TaskRunner.Run(ctx, taskPayload.Payload()); err != nil {
+		return err
+	}
+
+	a.UpdateArchiveExtractionStatus(lo.ToPtr(asset.ArchiveExtractionStatusInProgress))
+	if err := i.repos.Asset.Save(ctx, a); err != nil {
+		return err
+	}
+
+	if err := i.event(ctx, Event{
```
rot1024, line 170: move to original usecase function because it should be called even if decompression is skipped

rot1024 (changes requested):

rot1024 (approved):

PR #627, opened 2023-01-30
fix(server): request filtering is not working correctly

yk-eukarya:


rot1024 (approved):

rot1024:
It's better to create a struct type for arguments, but now ok

PR #633, opened 2023-02-02
feat(web): add boolean field

issmail-basel:


web/src/components/molecules/Content/Form/index.tsx
```diff
@@ -140,6 +142,8 @@ const ContentForm: React.FC<Props> = ({
       const values = await form.validateFields();
       const fields: { schemaFieldId: string; type: FieldType; value: string }[] = [];
       for (const [key, value] of Object.entries(values)) {
+        console.log(values);
```
nourbalaha, line 145: I think that this line should be deleted.

nourbalaha (approved):

PR #652, opened 2023-02-16
refactor(server): asset and local file system

yk-eukarya:


server/internal/usecase/interactor/asset.go
```diff
@@ -343,10 +326,10 @@ func (i *Asset) UpdateFiles(ctx context.Context, aid id.AssetID, s *asset.Archiv
 
 func detectPreviewType(files []gateway.FileEntry) *asset.PreviewType {
 	for _, entry := range files {
-		if path.Base(entry.Name) == "tileset.json" {
+		if filepath.Base(entry.Name) == "tileset.json" {
```
rot1024, line 329: Even on Windows, it would be better to normalize paths to path package's one instead of filepath's one outside `repo/fs.File`. If differences among operating systems come into play here, bugs will occur if implemented in a use case without knowing the specifications.

rot1024 (changes requested):

rot1024 (approved):

PR #677, opened 2023-03-20
feat(web): add pagination to request modal

issmail-basel:


web/src/components/organisms/Project/Content/hooks.ts
```diff
@@ -19,21 +19,32 @@ export default () => {
...
   const t = useT();
 
-  const { data: requestData } = useGetRequestsQuery({
+  const [page, setPage] = useState<number>(1);
+  const [pageSize, setPageSize] = useState<number>(10);
+
+  useEffect(() => {
+    setPage(page ? +page : 1);
+    setPageSize(pageSize ? +pageSize : 10);
+  }, [setPage, setPageSize, page, pageSize]);
```
nourbalaha, line 28: Just a question, I am not sure why you are checking the `page` and `pageSize` variables since they are initialized, how about this? just a suggestion
```suggestion
  useEffect(() => {
    page !== 1 && setPage(+page);
    pageSize !== 10 && setPageSize( +pageSize );
  }, [setPage, setPageSize, page, pageSize]);
  ```

nourbalaha (approved):

web/src/components/organisms/Project/Content/hooks.ts
issmail-basel: You are right, because other parts of code, pagination is dependent on query string and query string may be empty, here we don't need this check

PR #691, opened 2023-04-06
feat(server,web): asset upload with signed URL

orisano:
# Overview

## What I've done
Added the feature to upload directly to gcs by using a presigned url (to avoid timeouts for Cloud Run, etc.)
If the Store does not support presigned url, it will fallback to normal upload.


server/pkg/asset/upload.go
```diff
@@ -0,0 +1,32 @@
+package asset
+
+import (
+	"time"
+)
+
+type Upload struct {
+	uuid      string
+	project   ProjectID
```
rot1024, line 9:
```suggestion
	project   ProjectID
	user      UserID
```

It is better to add userID to Upload and use it on MongoDB query to filter uploads by userID

server/internal/infrastructure/fs/file.go
```diff
@@ -124,6 +125,14 @@ func (f *fileRepo) GetURL(a *asset.Asset) string {
 	return f.urlBase.JoinPath(assetDir, fileUUID[:2], fileUUID[2:], url.PathEscape(a.FileName())).String()
 }
 
+func (f *fileRepo) IssueUploadAssetLink(ctx context.Context, filename string, expiresAt time.Time) (string, string, error) {
+	return "", "", gateway.ErrUnsupportedOperation
```
rot1024, line 129: Considering that it is OSS and will run in a variety of environments, it would be great if the upload URL could be used for file systems as well. However, this could be done in a separate PR or implemented by someone else.

rot1024 (changes requested):

rot1024 (approved):

PR #698, opened 2023-04-24
feat(server): logical deletion for item

nourbalaha:
# Overview
Currently, if an item is deleted in CMS, it can never be restored. However, in the unlikely event that someone accidentally deletes the item, it will have to be restored from a database backup or re-registered. To avoid this, the item deletion will be changed to logical deletion, making it easier to restore the item from the MongoDB Atlas console if it is accidentally deleted.


yk-eukarya:
will be closed with #951

PR #723, opened 2023-05-22
feat(server): support AWS S3

nourbalaha:
# Overview
In this PR I implemented the `gateway.File` interface to support AWS S3 alongside GCS.


server/internal/infrastructure/aws/file.go
```diff
@@ -0,0 +1,263 @@
...
+func (f *fileRepo) read(ctx context.Context, filename string) (io.ReadCloser, error) {
+	if filename == "" {
+		return nil, rerror.ErrNotFound
+	}
+
+	buf := aws.NewWriteAtBuffer([]byte{})
+	_, err := f.s3Downloader.DownloadWithContext(ctx, buf, &s3.GetObjectInput{
+		Bucket: aws.String(f.bucketName),
+		Key:    aws.String(filename),
+	})
```
rot1024, line 190: The current implementation keeps all contents in memory. This means that simply trying to download a large file can easily run out of memory.

Instead of keeping it in memory, concatenate the streams and pass the data from S3 to the response.

```go
bucket := "your-bucket-name" 
key    := "path/to/key/filename.ext" 

resp, err := f.s3Client.GetObject(&s3.GetObjectInput{
    Bucket: aws.String(bucket),
    Key:    aws.String(key),
})
if err != nil {
	return rerror.ErrInternalBy(fmt.Errorf("s3: failed to get object: %w", err))
}

defer func() {
  _ = resp.Body.Close()
}()
return c.Stream(http.StatusOK, contentType, resp.Body)
```

server/internal/infrastructure/aws/file.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	return path.Join(s3AssetBasePath, uuid[:2], uuid[2:], objectName)
+}
+
+func newUUID() string {
+	return uuid.New().String()
+}
+
+func IsValidUUID(u string) bool {
```
rot1024, line 255:
```suggestion
func isValidUUID(u string) bool {
```

rot1024 (changes requested):

server/internal/infrastructure/aws/file.go
nourbalaha: Done!

server/internal/infrastructure/aws/file.go
nourbalaha: Done!

server/internal/infrastructure/aws/file.go
```diff
@@ -0,0 +1,276 @@
...
+		return nil, rerror.NewE(i18n.T("bucket name is empty"))
+	}
+
+	if secretAccessKey == "" || accessKeyID == "" {
+		return nil, rerror.NewE(i18n.T("no credentials provided"))
+	}
+
+	if region == "" {
+		return nil, rerror.NewE(i18n.T("region is empty"))
+	}
```
rot1024, line 49:
```suggestion
```

If the region is empty, `AWS_REGION` will be loaded automatically, so it should not return an error

server/internal/infrastructure/aws/file.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+func NewFile(bucketName, accessKeyID, secretAccessKey, region, baseURL, cacheControl string) (gateway.File, error) {
+	if bucketName == "" {
+		return nil, rerror.NewE(i18n.T("bucket name is empty"))
+	}
+
+	if secretAccessKey == "" || accessKeyID == "" {
+		return nil, rerror.NewE(i18n.T("no credentials provided"))
+	}
```
rot1024, line 45:
```suggestion
```

Crediential will be automatically loaded by AWS SDK for Go, so no need to return an error

Also note that this error is not displayed to the user and does not require translation.

server/internal/infrastructure/aws/file.go
```diff
@@ -0,0 +1,276 @@
...
+
+	var err error
+	u, _ = url.Parse(baseURL)
+	if err != nil {
+		return nil, rerror.NewE(i18n.T("invalid base URL"))
+	}
+
+	cfg, err := config.LoadDefaultConfig(context.TODO(),
+		config.WithRegion(region),
+		config.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(accessKeyID, secretAccessKey, "")),
```
rot1024, line 64:
```suggestion
```

https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/

As much as possible, leave it to the SDK to load the configuration automatically, and support different configuration methods such as environment variables and files. It can also load the configuration automatically with zero configuration when running in an AWS environment.

server/internal/infrastructure/aws/file.go
```diff
@@ -0,0 +1,276 @@
...
+type fileRepo struct {
+	bucketName   string
+	baseURL      *url.URL
+	cacheControl string
+	s3Client     *s3.Client
+}
+
+func NewFile(bucketName, accessKeyID, secretAccessKey, region, baseURL, cacheControl string) (gateway.File, error) {
+	if bucketName == "" {
+		return nil, rerror.NewE(i18n.T("bucket name is empty"))
```
rot1024, line 40:
```suggestion
		return nil, errors.New("bucket name is empty")
```

This error is not displayed to the user and does not require translation.

server/internal/infrastructure/aws/file.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	presignClient := s3.NewPresignClient(f.s3Client)
+	uploadURL, err := presignClient.PresignPutObject(ctx, &s3.PutObjectInput{
+		Bucket:       aws.String(f.bucketName),
+		CacheControl: aws.String(f.cacheControl),
+		Key:          aws.String(p),
+		ContentType:  aws.String(contentType),
+	}, func(options *s3.PresignOptions) {
+		options.Expires = time.Minute * 15
```
rot1024, line 162:
```suggestion
		options.Expires = expires
```

Define it as constant: `const expries = time.Minute * 15`

rot1024 (changes requested):

server/internal/infrastructure/aws/file.go
```diff
@@ -0,0 +1,276 @@
...
+)
+
+type fileRepo struct {
+	bucketName   string
+	baseURL      *url.URL
+	cacheControl string
+	s3Client     *s3.Client
+}
+
+func NewFile(bucketName, accessKeyID, secretAccessKey, region, baseURL, cacheControl string) (gateway.File, error) {
```
rot1024, line 38:
```suggestion
func NewFile(ctx context.Context, bucketName, accessKeyID, secretAccessKey, region, baseURL, cacheControl string) (gateway.File, error) {
```

pass context and use it with `config.LoadDefaultConfig`

rot1024 (changes requested):

server/internal/infrastructure/aws/file.go
nourbalaha: Done!

server/internal/infrastructure/aws/file.go
nourbalaha: Done!

server/internal/infrastructure/aws/file.go
nourbalaha: Done!

server/internal/infrastructure/aws/file.go
nourbalaha: Done!

server/internal/infrastructure/aws/file.go
nourbalaha: Done!

server/internal/infrastructure/aws/file.go
nourbalaha: Done!

PR #726, opened 2023-05-25
feat(worker): add AWS S3 and SNS support

nourbalaha:
# Overview
In this PR, I added support for AWS in the worker by implementing the FileRepo interface and creating the S3 and SNS files.


worker/internal/infrastructure/aws/sns.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+}
+
+func (s *SNS) NotifyAssetDecompressed(ctx context.Context, assetID string, status *asset.ArchiveExtractionStatus) error {
+	cfg, err := config.LoadDefaultConfig(ctx)
+	if err != nil {
+		return err
+	}
+
+	snsClient := sns.NewFromConfig(cfg)
```
rot1024, line 30:
```suggestion
```

worker/internal/infrastructure/aws/sns.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/reearth/reearth-cms/worker/pkg/asset"
+	"github.com/reearth/reearthx/log"
+)
+
+type SNS struct {
+	topicArn string
+}
+
+func NewSNS(topicArn string) *SNS {
+	return &SNS{
```
rot1024, line 19:
```suggestion

	cfg, err := config.LoadDefaultConfig(ctx)
	if err != nil {
		return err
	}

	snsClient := sns.NewFromConfig(cfg)
	
	return &SNS{
		snsClient: snsClient,
```

worker/internal/infrastructure/aws/s3.go
```diff
@@ -0,0 +1,148 @@
...
+	resp, err := f.s3Client.GetObject(ctx, params)
+	if err != nil {
+		log.Errorf("aws: read object err: %+v\n", err)
+		return nil, 0, 0, rerror.ErrInternalBy(err)
+	}
+
+	proceeded, _ := strconv.ParseInt(resp.Metadata["proceeded"], 10, 64)
+
+	// Read all data into memory
+	objectData, err := io.ReadAll(resp.Body)
```
rot1024, line 75: As I pointed out before, if you load a large file, the server will go down due to lack of memory in one shot.

It is a good idea to analyze the cause of the need for io.ReaderAt and improve it so that ReaderAt is not needed. If you are still not sure, please consult with us again.

rot1024 (changes requested):

worker/internal/infrastructure/aws/s3.go
rot1024: https://github.com/markandrus/s3readerat may be helpful

worker/internal/infrastructure/aws/sns.go
nourbalaha: Done!

worker/internal/infrastructure/aws/sns.go
nourbalaha: Done!

worker/internal/infrastructure/aws/s3.go
nourbalaha: It appears that the `readerAt` format is required as an input for the `archive/zip` package in order to read from a specific offset. 
Currently, our implementation on GCP follows a similar approach by reading everything into memory.

Could you please provide suggestions on how to fix the Read method for both AWS and GCP?

rot1024: Hmm current implementation is OK

We can implement a way to prevent loading the entire file at a later date. The simplest way is to write to a file. Then we can consider using:

- https://github.com/markandrus/s3readerat
- https://github.com/snabb/httpreaderat
- https://github.com/avvmoto/buf-readerat

rot1024 (approved):

worker/internal/infrastructure/aws/s3.go
nourbalaha: Well, that makes sense. I think that `buf-readerat` or a similar package could come in handy to solve this problem in the future.

Thanks!

PR #729, opened 2023-05-31
feat(worker): add AWS support to decompress and webhook handlers

nourbalaha:
# Overview
The decompress handler has been updated to support AWS SNS notifications
The webhook handler has been updated to support AWS SNS notifications


worker/internal/app/handler.go
```diff
@@ -39,16 +53,30 @@ func (h Handler) DecompressHandler() echo.HandlerFunc {
...
+			}
+			if err := json.Unmarshal([]byte(payload.Message), &w); err != nil {
+				return err
+			}
+			// Validates payload's signature
+			if err := payload.VerifyPayload(); err != nil {
+				return err
+			}
+		} else {
+			var msg msgBody
```
rot1024, line 70: If you intend to use PubSub, e.g. `else if isGCP() ...` is required. Also separate if else blocks into functions:

```go
func (h Handler) isSNS(r *http.Request) bool {

}

func (h Handler) isGCP(r *http.Request) bool {

}

func parseSNSMessage(body io.Reader) (webhook.Webhook, error) {

}

func parsePubSubMessage(body io.Reader) (webhook.Webhook, error) {

}
```

nourbalaha: Done!

worker/internal/app/handler.go
```diff
@@ -39,28 +52,114 @@ func (h Handler) DecompressHandler() echo.HandlerFunc {
...
+		return input, err
+	}
+	
+	return input, nil
+}
+
+func parsePubSubDecompressMessage(c echo.Context, body io.Reader) (rhttp.DecompressInput, error) {
+	var input rhttp.DecompressInput
+	
+	if err := c.Bind(&input); err != nil {
```
rot1024, line 118: Use `json.NewDecoder(body).Decode(&input)` instead of `c.Bind(&input)`

worker/internal/app/handler.go
```diff
@@ -39,28 +52,114 @@ func (h Handler) DecompressHandler() echo.HandlerFunc {
...
+	}
+
+	return w, nil
+}
+
+func parsePubSubWebhookMessage(c echo.Context, body io.Reader) (webhook.Webhook, error) {
+	var msg msgBody
+	var w webhook.Webhook
+
+	if err := c.Bind(&msg); err != nil {
```
rot1024, line 150: Use `json.NewDecoder(body).Decode(&input)` instead of `c.Bind(&input)`

worker/internal/app/handler.go
nourbalaha: I haven't wrote this code.
This is the original code that we have right now for GCP.
Done!

worker/internal/app/handler.go
nourbalaha: I haven't wrote this code.
This is the original code that we have right now for GCP.
Done!

rot1024 (approved):

PR #749, opened 2023-06-22
feat(server): enable/disable request feature for specific roles in projects

nourbalaha:
# Overview
In this PR I added `RequestRoles` slice to `Project` type.
The specified roles in the `RequestRoles` will not be able to publish without making requests.


server/pkg/project/builder.go
```diff
@@ -84,3 +86,8 @@ func (b *Builder) Publication(publication *Publication) *Builder {
 	b.p.publication = publication
 	return b
 }
+
+func (b *Builder) RequestRoles(requestRoles []user.Role) *Builder {
+	b.p.requestRoles = requestRoles
```
rot1024, line 91:
```suggestion
	b.p.requestRoles = slices.Clone(requestRoles)
```

I have made this same point sometimes. Do you understand why this is necessary?

server/pkg/project/project.go
```diff
@@ -93,6 +99,10 @@ func (p *Project) UpdateDescription(description string) {
 	p.description = description
 }
 
+func (p *Project) SetRequestRoles(sr []user.Role) {
+	p.requestRoles = sr
```
rot1024, line 103:
```suggestion
	p.requestRoles = slices.Clone(sr)
```

rot1024 (changes requested):

server/pkg/project/project.go
nourbalaha: Done!

server/pkg/project/builder.go
nourbalaha: Yes, I understand why this is necessary. I am sorry about forgetting to add `slices.Clone()`.
Done!

rot1024 (approved):

PR #758, opened 2023-06-27
fix(server): swapped maintaining projects and owning projects

nourbalaha:
# Overview
In this PR, the issue of swapped "maintaining projects" and "owning projects" has been fixed.


rot1024 (approved):
This is fine, but it might be better to modify the operatorProjects to stop returning multivalues and return structs, so that it is less likely to make the same mistake in the future.

PR #764, opened 2023-06-30
feat(server): add old item value

issmail-basel:


server/pkg/integrationapi/item.go
```diff
@@ -56,3 +58,58 @@ func NewItem(i *item.Item, s *schema.Schema, assets *AssetContext) Item {
...
+func NewItemWithPreviousValue(i *item.Item, o *item.Item, s *schema.Schema, assets *AssetContext) ItemWithPreviousValue {
+	fs := lo.FilterMap(i.Fields(), func(f *item.Field, _ int) (FieldWithPreviousValue, bool) {
+		if s == nil {
+			return FieldWithPreviousValue{}, false
+		}
+		sf := s.Field(f.FieldID())
+		if sf == nil {
+			return FieldWithPreviousValue{}, false
+		}
+		ii, _ := lo.Find(o.Fields(), func(field *item.Field) bool {
```
mimoham24, line 93: please remove the debugging code
no need for `;` in golang

server/pkg/integrationapi/item.go
```diff
@@ -56,3 +58,58 @@ func NewItem(i *item.Item, s *schema.Schema, assets *AssetContext) Item {
...
+			return field.FieldID() == f.FieldID()
+		})
+
+		return FieldWithPreviousValue{
+			Id:    f.FieldID().Ref(),
+			Type:  lo.ToPtr(ToValueType(f.Type())),
+			Value: lo.ToPtr(ToValues(f.Value(), sf.Multiple(), assets)),
+			OldValue: lo.ToPtr(ToValues(ii.Value(), sf.Multiple(), assets)),
+			Key:   util.ToPtrIfNotEmpty(sf.Key().String()),
+		}, true
```
mimoham24, line 105: this logic always returns the old value even if it's not changed 
for example: if we have two fields like:
```
{
        "type": "text",
        "value": "value 1"
},{
        "type": "text",
        "value": "value 2"
}
```
updating the first field value to `new value1` will have the output
```
{
        "type": "text",
        "value": "value 1",
        "oldValue": "new value1"
},{
        "type": "text",
        "value": "value 2",
        "oldValue": "value2"
}
```

mimoham24 (changes requested):

server/pkg/integrationapi/item.go
```diff
@@ -56,3 +58,58 @@ func NewItem(i *item.Item, s *schema.Schema, assets *AssetContext) Item {
...
+		if s == nil {
+			return FieldWithPreviousValue{}, false
+		}
+		sf := s.Field(f.FieldID())
+		if sf == nil {
+			return FieldWithPreviousValue{}, false
+		}
+		ii, _ := lo.Find(o.Fields(), func(field *item.Field) bool {
+			fmt.Println(field.FieldID().Ref());
+			fmt.Println( f.FieldID().Ref());
```
mimoham24, line 95: please remove the debugging code
no need for `;` in golang

mimoham24 (changes requested):

PR #773, opened 2023-07-06
fix(server): prevent empty options

nourbalaha:
# Overview
This fixes the bug where user was able to create a select field with empty options.
Now an error message will be displayed to prevent the user from creating a select field with empty options


server/i18n/ja.yml
```diff
@@ -1,6 +1,7 @@
 Comment already exist in this thread: コメントはすでにこのスレッドに存在します。
 Comment does not exist in this thread: コメントはこのスレッドに存在しません。
 File is missing: ファイルがありません
+Options could not be empty!: ""
```
HideBa, line 4:
```suggestion
Options could not be empty!: "選択フィールドは空にできません"
```

HideBa (approved):
kindly check transtations

PR #778, opened 2023-07-12
feat(web): add copy icon next to images' title

nourbalaha:
# Overview
- Add copy icon next to images' title


issmail-basel (approved):
LGTM!

PR #780, opened 2023-07-13
feat(server): add item change to webhook payload of item update event

issmail-basel:


server/pkg/integrationapi/schema.go
```diff
@@ -56,3 +73,74 @@ func NewSchema(i *schema.Schema) Schema {
 		CreatedAt: lo.ToPtr(i.ID().Timestamp()),
 	}
 }
+
+func NewFieldChangeType(value string) *FieldChangeType {
+	ft := FieldChangeType(value)
+	return &ft
+}
+
+func NewItemChange(n []*item.Field, o []*item.Field) []FieldChange {
```
rot1024, line 82: This logic should be put in item package. Also define FieldChange struct in item package. Comparing items to each other to get the difference is a fairly generic process that you may want to use in other situations in the future.

server/pkg/item/field_change.go
```diff
@@ -0,0 +1,85 @@
...
+
+type FieldChange struct {
+	Id *FieldID
+	CurrentValue any
+	PreviousValue any
+	Type FieldChangeType
+}
+
+
+func NewItemChange(n []*Field, o []*Field) []FieldChange {
```
rot1024, line 21:
```suggestion
func CompareFields(n []*Field, o []*Field) []FieldChange {
```

Also add unit tests for this func.

server/pkg/item/item.go
```diff
@@ -128,3 +128,11 @@ type ItemModelSchema struct {
 	Model  *model.Model
 	Schema *schema.Schema
 }
+
+type ItemModelSchemaItemChange struct {
```
rot1024, line 132: this is not used in item package. Move it to intergrationapi package.

server/pkg/item/field_change.go
```diff
@@ -0,0 +1,85 @@
...
+
+const (
+	Add    FieldChangeType = "add"
+	Update FieldChangeType = "update"
+	Delete FieldChangeType = "delete"
+)
+
+
+type FieldChange struct {
+	Id *FieldID
```
rot1024, line 14:
```suggestion
	ID *FieldID
```

In golang, abbreviations that are normally written in capital letters should be left as they are.

server/pkg/item/field_change.go
```diff
@@ -0,0 +1,85 @@
...
+	CurrentValue any
+	PreviousValue any
```
rot1024, line 16:
```suggestion
	CurrentValue *value.Multiple
	PreviousValue *value.Multiple
```

server/pkg/item/item.go
issmail-basel: @rot1024 , it is the exactly same as `ItemModelSchema` usage, which exists in the item package.

Also, this struct is used in `internal/usecase/interactor/item.go`, are you sure this should be moved?

server/pkg/item/field_change.go
issmail-basel: @rot1024 , I changed it, but it represents the hook return data so now the hook returns:

```
 "fieldschange": [
      {
        "CurrentValue": [
          "added value"
        ],
        "ID": "01h5v15gfkrerr0zrqa914zan6",
        "PreviousValue": null,
        "Type": "add"
      }
    ],
```

server/pkg/item/field_change.go
issmail-basel: @rot1024 , by the way this one already has `any` type in `internal/adapter/gql/gqlmodel/convert_value.go`

```
func ToValue(v *value.Multiple, multiple bool) any {
	if !multiple {
		return v.First().Interface()
	}
	return v.Interface()
}
```

Changed it to:

```
	CurrentValue []interface{}
	PreviousValue []interface{}
```
as it is an interface of that value.

server/pkg/item/field_change.go
rot1024:
```go
type A struct {
	ID string `json:"id"`
}
```

You can use "tag". For details, google it

server/pkg/item/field_change.go
rot1024: This struct cannot be used as is in the GraphQL world.In the domain model layer, the domain model information should not be stripped away unnecessarily. Instead of using it in the GraphQL layer, process it as a domain model in the use case layer and convert it to GraphQL.

server/pkg/item/item.go
rot1024: I see, then it should rather be defined in the use case layer. If the use case is implemented to return that struct, it should be defined in the use case layer.

server/pkg/item/field_change_test.go
```diff
@@ -0,0 +1,142 @@
...
+		n []*Field
+		o []*Field
+	}
+	tests := []struct {
+		name string
+		args args
+		want []FieldChange
+	}{
+		{
+			name: "NoChanges",
```
rot1024, line 27: "no change"

server/pkg/item/field_change_test.go
```diff
@@ -0,0 +1,142 @@
...
+				},
+				o: []*Field{
+					NewField(fId, value.TypeText.Value("value1").AsMultiple()),
+					NewField(fId2, value.TypeNumber.Value("value1").AsMultiple()),
+				},
+			},
+			want: nil, // No changes expected
+		},
+			{
+			name: "AddField",
```
rot1024, line 41: "add"

server/pkg/item/field_change_test.go
```diff
@@ -0,0 +1,142 @@
...
+				{
+					ID:            &fId2,
+					Type:          Add,
+					CurrentValue:  []interface{}{"new field"},
+					PreviousValue: nil,
+				},
+			},
+		},
+		{
+			name: "UpdateField",
```
rot1024, line 61: "update"

server/pkg/item/field_change_test.go
```diff
@@ -0,0 +1,142 @@
...
+				{
+					ID:            &fId,
+					Type:          Update,
+					PreviousValue: []interface{}{"value1"},
+					CurrentValue:  []interface{}{"value2"},
+				},
+			},
+		},
+		{
+			name: "DeleteField",
```
rot1024, line 82: "delete"

server/pkg/item/item.go
issmail-basel: @rot1024 , after much thinking, I found out that I don't need to create a separate struct, I used existing `ItemModelSchema` with additional optional properties as needed.

rot1024 (approved):

PR #781, opened 2023-07-20
feat(server): add checkIfItemIsReferenced query

nourbalaha:
# Overview
- Add `checkIfItemIsReferenced` query


server/schemas/item.graphql
```diff
@@ -118,19 +118,21 @@ input ItemQuery {
 }
 
 extend type Query {
-  items(schemaId: ID!, sort: ItemSort, pagination: Pagination): ItemConnection!
+  items(modelId: ID!, pagination: Pagination): ItemConnection!
```
rot1024, line 121: where was sort gone?

server/schemas/field.graphql
```diff
@@ -27,6 +27,7 @@ type SchemaField {
   multiple: Boolean!
   unique: Boolean!
   required: Boolean!
+  isTitle: Boolean
```
rot1024, line 30:
```suggestion
  isTitle: Boolean!
```

I am concerned that only this field has Is in the name, but it would be confusing to name it title, so this is fine.

server/schemas/item.graphql
```diff
@@ -68,8 +68,8 @@ input UnpublishItemInput {
   itemId: [ID!]!
 }
 
-input publishItemInput {
-  itemId: ID!
+input PublishItemInput {
+  itemId: [ID!]!
```
rot1024, line 72:
```suggestion
  itemIds: [ID!]!
```

server/schemas/field.graphql
```diff
@@ -97,6 +98,8 @@ type SchemaFieldInteger {
 
 type SchemaFieldReference {
   modelId: ID!
+  correspondingField: SchemaField
+  correspondingFieldId: ID
```
rot1024, line 102:
```suggestion
  correspondingFieldId: ID!
```

required?

server/schemas/item.graphql
```diff
@@ -118,19 +118,21 @@ input ItemQuery {
 }
 
 extend type Query {
-  items(schemaId: ID!, sort: ItemSort, pagination: Pagination): ItemConnection!
+  items(modelId: ID!, pagination: Pagination): ItemConnection!
+  itemsByIds(ids: [ID!]!): ItemConnection!
```
rot1024, line 122:
```suggestion
```

`nodes` query is available. Use it.

server/schemas/item.graphql
```diff
@@ -118,19 +118,21 @@ input ItemQuery {
...
   versionsByItem(itemId: ID!): [VersionedItem!]!
   searchItem(
     query: ItemQuery!
     sort: ItemSort
     pagination: Pagination
   ): ItemConnection!
+  checkIfItemIsReferenced(itemId: ID!): Boolean
```
rot1024, line 129: Please give me a use case. This query seems to have a very limited use case. I think it could be made into a more useful query.

server/pkg/schema/field.go
```diff
@@ -20,6 +20,7 @@ type Field struct {
 	unique       bool
 	multiple     bool
 	required     bool
+	isTitle      *bool
```
rot1024, line 23:
```suggestion
	isTitle      bool
```

The field should have no state other than being titled or not titled.

server/pkg/schema/field.go
```diff
@@ -101,6 +102,10 @@ func (f *Field) Required() bool {
 	return f.required
 }
 
+func (f *Field) IsTitle() *bool {
```
rot1024, line 105:
```suggestion
func (f *Field) IsTitle() bool {
```

server/pkg/schema/field.go
```diff
@@ -113,6 +118,10 @@ func (f *Field) SetMultiple(m bool) {
 	f.multiple = m
 }
 
+func (f *Field) SetIsTitle(isTitle *bool) {
```
rot1024, line 121:
```suggestion
func (f *Field) SetIsTitle(isTitle bool) {
```

server/internal/usecase/interfaces/schema.go
```diff
@@ -20,6 +20,7 @@ type CreateFieldParam struct {
 	Multiple     bool
 	Unique       bool
 	Required     bool
+	IsTitle      *bool
```
rot1024, line 23:
```suggestion
	IsTitle      bool
```

server/internal/usecase/repo/schema.go
```diff
@@ -11,6 +11,7 @@ type Schema interface {
 	Filtered(filter WorkspaceFilter) Schema
 	FindByIDs(context.Context, id.SchemaIDList) (schema.List, error)
 	FindByID(context.Context, id.SchemaID) (*schema.Schema, error)
+	FindFieldByIDs(context.Context, id.FieldIDList) ([]*schema.Field, error)
```
rot1024, line 14:
```suggestion
```

It is not permitted to take only a portion of the aggregate out of the repository. Recall the definition of an aggregate.

More to the point, looking up a schema by field ID is very inefficient; it increases the DB indexes. There is an easier way. When you create some reference field for a schema, you can make it have the ID of the schema it refers to.

rot1024 (changes requested):
Next time, please split the pull request into smaller pieces.

Allowing multiple items to be published at once and implementing a reference field should be separate implementations.

server/internal/usecase/interfaces/schema.go
nourbalaha: Done!

server/pkg/schema/field.go
nourbalaha: Done!

server/schemas/field.graphql
nourbalaha: I think that this field should be optional because it is only required in the two-way reference fields!

server/pkg/schema/field.go
nourbalaha: Done!

server/pkg/schema/field.go
nourbalaha: Done!

server/schemas/item.graphql
nourbalaha: I am using this to check if an item is reference or not by id. 
This is useful in the frontend to give the user a warning in case of dereferencing a referenced item.
Please let me know if you have a better way to do so.

server/schemas/item.graphql
nourbalaha: Done!

server/internal/usecase/repo/schema.go
nourbalaha: Just to clerify any misunderstanding, I am getting fields by fieldIds.
I am using this for the corresponding field resolver
```
  SchemaFieldReference:
    fields:
      correspondingField:
        resolver: true
```
Are you suggesting that I should remove the resolver and instead I should save the schema Id??

server/schemas/item.graphql
nourbalaha: `FindByModel` method doesn't have sort in the usecase layer, I added sort to all layers.

server/schemas/item.graphql
nourbalaha: Removed it, done!

PR #798, opened 2023-08-10
feat(server): account database to be configurable

orisano:
# Overview


server/internal/app/config.go
```diff
@@ -49,6 +49,8 @@ type Config struct {
 	Auth_JWKSURI  *string
 	// auth for m2m
 	AuthM2M AuthM2MConfig
+
+	AccountDatabase string
```
rot1024, line 53:
```suggestion
	DB_Account string
```

rot1024 (approved):

PR #801, opened 2023-08-23
feat(server): meta schema fields

mimoham24:


server/internal/usecase/interfaces/model.go
```diff
@@ -39,6 +38,7 @@ type Model interface {
 	FindByProject(context.Context, id.ProjectID, *usecasex.Pagination, *usecase.Operator) (model.List, *usecasex.PageInfo, error)
 	FindByKey(context.Context, id.ProjectID, string, *usecase.Operator) (*model.Model, error)
 	FindByIDOrKey(context.Context, id.ProjectID, model.IDOrKey, *usecase.Operator) (*model.Model, error)
+	CreateMetadata(context.Context, id.ModelID, *usecase.Operator) (*model.Model, error)
```
rot1024, line 41: is it used?

server/internal/usecase/interactor/item.go
```diff
@@ -304,6 +305,19 @@ func (i Item) Update(ctx context.Context, param interfaces.UpdateItemParam, oper
 		oldFields := itv.Fields()
 
 		itv.UpdateFields(fields)
+
+		if operator.AcOperator.User != nil {
+			itv.SetUpdatedByUser(*operator.AcOperator.User)
+		}
+
+		if operator.Integration != nil {
```
rot1024, line 313:
```suggestion
		} else if operator.Integration != nil {
```

server/internal/usecase/interactor/item.go
```diff
@@ -208,6 +208,7 @@ func (i Item) Create(ctx context.Context, param interfaces.CreateItemParam, oper
 			Schema(s.ID()).
 			Project(s.Project()).
 			Model(m.ID()).
+			MetadataItem(param.MetadataID).
```
rot1024, line 211: MetadaID seems not validated

server/internal/usecase/interactor/model.go
```diff
@@ -69,6 +69,10 @@ func (i Model) Create(ctx context.Context, param interfaces.CreateModelParam, op
 			if err != nil {
 				return nil, err
 			}
+			meta, err := schema.New().NewID().Workspace(p.Workspace()).Project(p.ID()).Build()
```
rot1024, line 72: Do you always create metadata? Not all models use metadata, so it is fine to create it on a delayed basis (when fields are added to the metadata for the first time).

server/internal/adapter/gql/resolver_mutation_item.go
```diff
@@ -23,10 +23,14 @@ func (r *mutationResolver) CreateItem(ctx context.Context, input gqlmodel.Create
 	if err != nil {
 		return nil, err
 	}
+	if err != nil {
+		return nil, err
+	}
```
rot1024, line 28:
```suggestion

```

server/internal/adapter/gql/resolver_mutation_field.go
```diff
@@ -68,18 +67,25 @@ func (r *mutationResolver) UpdateField(ctx context.Context, input gqlmodel.Updat
...
+		s, err = usecases(ctx).Schema.FindByID(ctx, *m.Metadata(), getOperator(ctx))
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		s, err = usecases(ctx).Schema.FindByID(ctx, m.Schema(), getOperator(ctx))
+		if err != nil {
+			return nil, err
+		}
 	}
```
rot1024, line 89: These logic should be moved to the usecase layer. This is the adapter layer, so it should be thin.

server/internal/usecase/interactor/model.go
rot1024: Is it possible to add fields to the metadata without the metadata existing in the schema? Does an implementation exist?

mimoham24: metadata is created on field creation for the first time

server/i18n/ja.yml
```diff
@@ -77,4 +77,5 @@ unsupported operation: サポートされていない処理です。
 uuid is required: UUIDは必須です。
 value is required: 値は必須です。
 workspace id is required: ワークスペースIDは必須です。
-metadata schema not found: ""
\ No newline at end of file
+metadata schema not found: ""
+metadata item and schema mismatch: ""
```
rot1024, line 81:
```suggestion
metadata schema not found: メタデータのスキーマが見つかりません。
metadata item and schema mismatch: メタデータのアイテムのスキーマが正しくありません。
```

server/internal/adapter/gql/resolver_mutation_field.go
```diff
@@ -35,7 +43,7 @@ func (r *mutationResolver) CreateField(ctx context.Context, input gqlmodel.Creat
...
```
rot1024, line 43: line 22~43 should be done in the usecase layer. CreateField usecase should accept modelID and metadata bool. The adapter layer should be thin as possible.

mimoham24: do you mean line 22~39?  lines 40~43 are using a gqlmodel convertor and we are following the same logic everywhere, If so we need to move this logic from the update function as well

server/internal/usecase/interfaces/model.go
```diff
@@ -20,6 +20,11 @@ type CreateModelParam struct {
 	Key         *string
 	Public      *bool
 }
+type FindOrCreateSchemaParam struct {
+	ModelId  id.ModelID
+	Metadata *bool
+	Create   bool
```
rot1024, line 26:
```suggestion
	ModelID  id.ModelID
	// boolean that identify if it is a metdata
	Metadata *bool
	// boolean to identify if we want to create a metadata schema or just return an error if metadata schema is nil
	Create   bool
```

server/internal/usecase/interfaces/model.go
```diff
@@ -20,6 +20,11 @@ type CreateModelParam struct {
 	Key         *string
 	Public      *bool
 }
+type FindOrCreateSchemaParam struct {
```
rot1024, line 23:
```suggestion

type FindOrCreateSchemaParam struct {
```

insert line break between struct

server/internal/usecase/interfaces/model.go
```diff
@@ -20,6 +20,11 @@ type CreateModelParam struct {
...
+	ModelId  id.ModelID
+	Metadata *bool
+	Create   bool
+}
 
 type UpdateModelParam struct {
 	ModelId     id.ModelID
```
rot1024, line 30:
```suggestion
	ModelID     id.ModelID
```

rot1024 (approved):

PR #804, opened 2023-08-24
feat(server): add titleField to schema

nourbalaha:
# Overview
- Add `TitleField` to `Schema`


server/e2e/publicapi_test.go
```diff
@@ -270,12 +270,13 @@ func publicAPISeeder(ctx context.Context, r *repo.Container) error {
 		FileName("aaa.zip").UUID(publicAPIAssetUUID).MustBuild()
 	af := asset.NewFile().Name("bbb.txt").Path("aaa/bbb.txt").Build()
 
+	fId := id.NewFieldID()
```
rot1024, line 273:
```suggestion
	fid := id.NewFieldID()
```

In golang, `Id` is not used as a variable name and a field name. Use `id` or `ID` instead.

See https://github.com/golang/go/wiki/CodeReviewComments#initialisms

> Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url". As an example: ServeHTTP not ServeHttp. For identifiers with multiple initialized "words", use for example "xmlHTTPRequest" or "XMLHTTPRequest".
> 
> This rule also applies to "ID" when it is short for "identifier" (which is pretty much all cases when it's not the "id" as in "ego", "superego"), so write "appID" instead of "appId".
>
> Code generated by the protocol buffer compiler is exempt from this rule. Human-written code is held to a higher standard than machine-written code.

e.g.

```go
type Hoge struct {
  aId string // ng
  aID string // ok
  aid string // ok
}
```

server/pkg/schema/schema.go
```diff
@@ -77,15 +82,29 @@ func (s *Schema) RemoveField(fid FieldID) {
 	}
 }
 
+func (s *Schema) TitleField() *FieldID {
+	return s.titleField
```
rot1024, line 86:
```suggestion
	return s.titleField.CloneRef()
```

```go
*s.TitleField() = "you can break fields inside the struct"
```

server/pkg/schema/schema.go
```diff
@@ -77,15 +82,29 @@ func (s *Schema) RemoveField(fid FieldID) {
...
+}
+
+func (s *Schema) SetTitleField(tf *FieldID) error {
+	if !slices.Contains(s.Fields().IDs(), *tf) {
+		return titleFieldErr
+	}
+
+	s.titleField = tf
```
rot1024, line 94:
```suggestion
	s.titleField = tf.CloneRef()
```

server/schemas/schema.graphql
```diff
@@ -2,6 +2,7 @@ type Schema implements Node {
   id: ID!
   projectId: ID!
   fields: [SchemaField!]!
+  titleField: ID
```
rot1024, line 5: You can also add `isTitle: Boolean!` to SchemaField

server/pkg/schema/builder.go
```diff
@@ -55,3 +55,8 @@ func (b *Builder) Fields(fields FieldList) *Builder {
 	b.s.fields = fields.Clone()
 	return b
 }
+
+func (b *Builder) TitleField(fId *FieldID) *Builder {
+	b.s.titleField = fId.CloneRef()
```
rot1024, line 60:
```suggestion
func (b *Builder) TitleField(fid *FieldID) *Builder {
	b.s.titleField = fid.CloneRef()
```

Also you forgot titleField validation in the `Build` method.

server/internal/usecase/interactor/schema.go
```diff
@@ -91,6 +97,13 @@ func (i Schema) UpdateField(ctx context.Context, param interfaces.UpdateFieldPar
 		if err := updateField(param, f); err != nil {
 			return nil, err
 		}
+
+		if param.IsTitle != nil && *param.IsTitle {
```
rot1024, line 101:
```suggestion
		if param.IsTitle != nil {
```

Currently there is no way to delete the title field from the schema field.

server/internal/infrastructure/mongo/mongodoc/schema.go
```diff
@@ -157,6 +159,7 @@ func (d *SchemaDocument) Model() (*schema.Schema, error) {
 	if err != nil {
 		return nil, err
 	}
+	fId := id.FieldIDFromRef(d.TitleField)
```
rot1024, line 162:
```suggestion
	fid := id.FieldIDFromRef(d.TitleField)
```

rot1024 (changes requested):

server/e2e/publicapi_test.go
nourbalaha: Okay so it is either fid or fID.
Done!

server/schemas/schema.graphql
nourbalaha: Done!

server/pkg/schema/schema.go
nourbalaha: Done!

server/pkg/schema/schema.go
nourbalaha: Done!

server/pkg/schema/builder.go
nourbalaha: TitleField is optional, what kind of validation is required in the build method?

server/internal/infrastructure/mongo/mongodoc/schema.go
nourbalaha: Done!

server/internal/usecase/interactor/schema.go
nourbalaha: Done!

server/pkg/schema/builder.go
nourbalaha: Done!

server/schemas/schema.graphql
nourbalaha: Done!

server/pkg/schema/schema.go
nourbalaha: Done!

server/pkg/schema/schema.go
nourbalaha: Done!

server/e2e/publicapi_test.go
nourbalaha: Okay, so fid and fID are acceptable but fId is not, understood._
Done!

server/schemas/schema.graphql
rot1024, line 5:
```suggestion
  titleFieldId: ID
  titleField: SchemaField
```

titleField is a resolver

rot1024 (approved):

server/schemas/schema.graphql
nourbalaha: Done!

PR #811, opened 2023-08-31
feat(server): update integration + webhook for ref field

nourbalaha:
# Overview
Update integration api and webhook for reference field


server/pkg/item/item.go
```diff
@@ -124,8 +124,9 @@ func (i *Item) AssetIDs() AssetIDList {
 }
 
 type ItemModelSchema struct {
-	Item    *Item
-	Model   *model.Model
-	Schema  *schema.Schema
-	Changes FieldChanges
+	Item           *Item
+	ReferencedItem *Item
```
rot1024, line 128: []*Item

nourbalaha: Done!

rot1024 (approved):

PR #812, opened 2023-09-04
feat(server): add isItemReferenced query

nourbalaha:
# Overview
- Add `isItemReferenced` query to check if an item is referenced by any other item

### DIFF
--- a/server/internal/usecase/interactor/item.go
+++ b/server/internal/usecase/interactor/item.go
@@ -160,6 +160,34 @@ func (i Item) Search(ctx context.Context, q *item.Query, sort *usecasex.Sort, p 
 	return i.repos.Item.Search(ctx, q, sort, p)
 }
 
+func (i Item) CheckIfItemIsReferenced(ctx context.Context, itemId id.ItemID, _ *usecase.Operator) (*bool, error) {
+	itm, err := i.repos.Item.FindByID(ctx, itemId, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	s, err := i.repos.Schema.FindByID(ctx, itm.Value().Schema())
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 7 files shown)


server/internal/usecase/interactor/item.go
```diff
@@ -160,6 +160,31 @@ func (i Item) Search(ctx context.Context, q *item.Query, sort *usecasex.Sort, p
 	return i.repos.Item.Search(ctx, q, sort, p)
 }
 
+func (i Item) CheckIfItemIsReferenced(ctx context.Context, itemId id.ItemID, correspondingFieldId id.FieldID, _ *usecase.Operator) (*bool, error) {
```
rot1024, line 163:
```suggestion
func (i Item) CheckIfItemIsReferenced(ctx context.Context, itemId id.ItemID, correspondingFieldID id.FieldID, _ *usecase.Operator) (bool, error) {
```

There is reason to use a pointer bool

server/schemas/item.graphql
```diff
@@ -125,6 +125,7 @@ extend type Query {
     sort: ItemSort
     pagination: Pagination
   ): ItemConnection!
+  checkIfItemIsReferenced(itemId: ID!, correspondingFieldId: ID!): Boolean
```
rot1024, line 128:
```suggestion
  checkIfItemIsReferenced(itemId: ID!, correspondingFieldId: ID!): Boolean!
```

nourbalaha: Done!

server/internal/usecase/interactor/item.go
nourbalaha: I wanted to return nil in case of an error, I will return false.
Done!

rot1024 (approved):

PR #817, opened 2023-09-04
feat(server): reference items

nourbalaha:
# Overview
- Implement reference items


server/pkg/item/item.go
```diff
@@ -69,6 +70,22 @@ func (i *Item) Thread() ThreadID {
...
+		return f.FieldID() == *fid
+	})
+	if !ok {
+		return
+	}
+	if v.Type() != i.Field(*fid).Type() {
+		return
+	}
+	i.fields[ii].value = v
+}
```
rot1024, line 87: This method is not needed I think

```go
i.UpdateFields([]*Field{item.NewField(fid, v)})
```

Basically, the private fields of a struct should only be accessible by that struct's methods. Accessing it from any other way will make the code very difficult to understand and will create bugs.

Also note that item.Field is an immtuable struct because there is no setter. However, this implementation breaks the immutability of the field.

nourbalaha: Understood, Thanks for the comment!
Done!

nourbalaha: Removed, Thanks for the comment!
Done!

server/pkg/item/reference.go
```diff
@@ -0,0 +1,38 @@
+package item
+
+import (
+	"github.com/reearth/reearth-cms/server/pkg/id"
+	"github.com/reearth/reearth-cms/server/pkg/schema"
+	"github.com/reearth/reearth-cms/server/pkg/value"
+)
+
+func AreItemsReferenced(i1, i2 *Item, s1, s2 *schema.Schema) (*id.FieldID, *id.FieldID, bool) {
```
rot1024, line 9:
```suggestion
func AreItemsReferenced(i1, i2 *Item, s1, s2 *schema.Schema) (*id.FieldID, *id.FieldID) {
```

nourbalaha: Done!

server/internal/usecase/interactor/item.go
```diff
@@ -287,6 +291,42 @@ func (i Item) Create(ctx context.Context, param interfaces.CreateItemParam, oper
...
+		}
+		s2, err := i.repos.Schema.FindByID(ctx, itm2.Value().Schema())
+		if err != nil {
+			return err
+		}
+		_, fid2, ok := item.AreItemsReferenced(it, itm2.Value(), s, s2)
+		if !ok {
+			continue
+		}
+		vv := value.New(value.TypeReference, it.ID().String()).AsMultiple()
```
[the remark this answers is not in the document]
nourbalaha: Done!

rot1024 (approved):

PR #854, opened 2023-09-19
feat(server): group and Item group field GQL schemas

mimoham24:


rot1024 (commented):
CreateFieldInput should be include `groupId: ID` (note it's optional) to enable users to add fields to a group.

server/schemas/item.graphql
```diff
@@ -27,12 +27,27 @@ type Item implements Node {
...
+  id: ID!
+  groupId: ID!
+  schemaId: ID!
+  modelId: ID!
+  projectId: ID!
+  schema: Schema!
+  group: Group!
+  model: Model!
+  project: Project!
+  fields: [ItemField!]!
```
rot1024, line 48:
```suggestion
  schemaFieldId: ID!
  groupId: ID!
  fields: [ItemField!]!
```

ItemValueField is simple, but this is too complex. Also this does not implements Node, so it can be simplified.

PR #855, opened 2023-09-19
feat(server): group domain models and GQL

mimoham24:


server/pkg/value/string.go
```diff
@@ -14,6 +14,7 @@ const TypeRichText Type = "richText"
 const TypeMarkdown Type = "markdown"
 const TypeSelect Type = "select"
 const TypeTag Type = "tag"
+const TypeGroup Type = "group"
```
rot1024, line 17: GroupID is not a string, it's id.GroupID. Refer to https://github.com/reearth/reearth-cms/blob/group/domain-models/server/pkg/value/reference.go#L12C4-L12C27

server/schemas/group.graphql
```diff
@@ -0,0 +1,47 @@
+type Group implements Node {
+    id: ID!
+    schemaId: ID!
+    projectId: ID!
+    name: String!
+    description: String!
+    key: String!
```
yk-eukarya, line 7: is this key required? is not schema field key enough?

mimoham24: yes sure, If you check the Figma the group should have a key field in a similar way to the model
also, the group schema field is not a group.... the schema group field is pointing at the group but they are both different domains

server/schemas/field.graphql
```diff
@@ -227,10 +240,12 @@ input SchemaFieldTypePropertyInput @onlyOne {
   integer: SchemaFieldIntegerInput
   reference: SchemaFieldReferenceInput
   url: SchemaFieldURLInput
+  group: SchemaFieldGroupInput
 }
 
 input CreateFieldInput {
-  modelId: ID!
+  modelId: ID
+  groupId: ID
```
yk-eukarya, line 248:
```suggestion
  schemaId: ID
```

server/schemas/group.graphql
yk-eukarya: >  the group schema field is not a group.... the schema group field is pointing at the group but they are both different entities

this is clear to me, but how we will use the group key?

server/schemas/field.graphql
mimoham24: I totally agree, but this change requires a lot of changes in our current implementation. Unfortunately, we have a critical deadline this time
I'm planning to refactor this part later

server/e2e/gql_field_test.go
```diff
@@ -21,7 +21,7 @@ func createField(e *httpexpect.Expect, mID, title, desc, key string, multiple, u
 				  }
 				}`,
 		Variables: map[string]any{
-			"modelId":      mID,
+			"modelId":      &mID,
```
yk-eukarya, line 24:
```suggestion
			"modelId":      mID,
```

server/e2e/gql_field_test.go
```diff
@@ -58,7 +58,7 @@ func updateField(e *httpexpect.Expect, mID, fID, title, desc, key string, multip
...
```
yk-eukarya:
```suggestion
			"modelId":      mID,
```

yk-eukarya (approved):
LGTM!

PR #863, opened 2023-09-25
feat(server): group usecase layer

mimoham24:


server/internal/usecase/interactor/group.go
```diff
@@ -0,0 +1,158 @@
...
+			}
+
+			if param.Name != nil {
+				g.SetName(*param.Name)
+			}
+			if param.Description != nil {
+				g.SetDescription(*param.Description)
+			}
+			if param.Key != nil {
+				if err := g.SetKey(key.New(*param.Key)); err != nil {
```
yk-eukarya, line 114: we should check if the key is still uniqu

server/internal/usecase/interactor/group.go
```diff
@@ -0,0 +1,158 @@
...
+			g, err := i.repos.Group.FindByKey(ctx, pId, s)
+			if g == nil && err == nil || err != nil && errors.Is(err, rerror.ErrNotFound) {
+				return true, nil
+			}
+
+			return false, err
+		})
+}
+
+func (i Group) Delete(ctx context.Context, groupID id.GroupID, operator *usecase.Operator) error {
```
yk-eukarya, line 142: is it OK to remove the group if it's used in some models?

server/internal/usecase/interactor/item.go
```diff
@@ -347,7 +355,12 @@ func (i Item) Update(ctx context.Context, param interfaces.UpdateItemParam, oper
...
-		fields, err := itemFieldsFromParams(param.Fields, s)
+		modelSchemaFields, otherFields := filterFieldParamsBySchema(param.Fields, s)
+
+		fields, err := itemFieldsFromParams(modelSchemaFields, s)
+		if err != nil {
+			return nil, err
+		}
 		if err != nil {
 			return nil, err
 		}
```
yk-eukarya, line 366:
```suggestion
		fields, err := itemFieldsFromParams(modelSchemaFields, s)
		if err != nil {
			return nil, err
		}
```

server/internal/usecase/interactor/item.go
```diff
@@ -696,12 +715,70 @@ func (i Item) handleReferenceFields(ctx context.Context, s schema.Schema, it *it
...
+			return nil, interfaces.ErrInvalidField
+		}
+
+		groupItemParams := lo.Filter(params, func(param interfaces.ItemFieldParam, _ int) bool {
+			if param.ItemGroup == nil {
+				return false
+			}
+			for _, groupValue := range mvg {
+				return groupValue == *param.ItemGroup
+			}
```
yk-eukarya, line 750: in this way, only the first item will be checked! use `lo.Find` instead

server/internal/usecase/interactor/group.go
```diff
@@ -149,10 +158,49 @@ func (i Group) Delete(ctx context.Context, groupID id.GroupID, operator *usecase
...
+	var schemas id.SchemaIDList
+	for _, m := range models {
+		msMap[m.Schema()] = m
+		schemas.Add(m.Schema())
+	}
+
+	sl, err := i.repos.Schema.FindByIDs(ctx, schemas)
+	if err != nil {
+		return nil, err
+	}
```
yk-eukarya, line 190:
```suggestion
	msMap := make(map[id.SchemaID]*model.Model)
	for _, m := range models {
		msMap[m.Schema()] = m
	}

	sl, err := i.repos.Schema.FindByIDs(ctx, lo.Keys(msMap))
	if err != nil {
		return nil, err
	}
```

yk-eukarya (approved):

PR #878, opened 2023-10-10
fix(server): add group to node type

nourbalaha:
# Overview
- `Group` has been added to `NodeType` enum


yk-eukarya (commented):
it seems you forget to implement the changes in the `Nodes` resolver

yk-eukarya (approved):

PR #890, opened 2023-10-17
feat(web): add filter component

caichi-t:
# Overview
Completed creation of filter component.
Throwing new searchItem API with filter conditions.
## Memo
Added filter conditions to URL but it may be better to encode the texts or do something because the URL may be too long.


web/src/components/molecules/Content/Table/DropdownRender.tsx
```diff
@@ -0,0 +1,295 @@
...
+            operatorType: "string",
+            value: StringOperator.NotEndsWith,
+            label: t("doesn't end with"),
+          },
+          { operatorType: "nullable", value: NullableOperator.Empty, label: t("is empty") },
+          { operatorType: "nullable", value: NullableOperator.NotEmpty, label: t("is not empty") },
+        );
+        break;
+      case "Integer":
+      case "Flaot":
```
nourbalaha, line 114:
```suggestion
      case "Float":
```

web/src/components/molecules/Content/Table/types.ts
```diff
@@ -1,2 +1,34 @@
...
 export type StateType = "DRAFT" | "PUBLIC" | "REVIEW";
+export type DefaultFilterValueType = {
+  operatorType: string;
+  operator: Operator;
+  value: string;
+};
+export type DropdownFilterType = {
+  dataIndex: string | string[];
+  title: string;
+  type: string;
```
nourbalaha, line 21: a type should be created here
```suggestion
  type: FilterType;
```

web/src/components/molecules/Content/Table/DropdownRender.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+  if (defaultValue) {
+    filterOption.current = {
+      value: defaultValue.operator,
+      operatorType: defaultValue.operatorType,
+    };
+  }
+  const filterValue = useRef<string>();
+
+  const confirm = () => {
```
nourbalaha, line 186:
```suggestion
  const confirm = useCallback(() => {
```

web/src/components/molecules/Content/Table/DropdownRender.tsx
```diff
@@ -0,0 +1,295 @@
...
+    }
+  } else if (filter.type === "Person") {
+    if (filter?.members?.length) {
+      for (const member of Object.values(filter.members)) {
+        valueOptions.push({ value: member.user.name, label: member.user.name });
+      }
+    }
+  } else if (filter.type === "Bool") {
+    valueOptions.push({ value: "true", label: "True" }, { value: "false", label: "False" });
+  }
```
nourbalaha, line 174:
```suggestion
  const valueOptions = useMemo<{
    value: string;
    label: string;
  }>(() => {
    const options = [];

    if (filter.type === "Select") {
      if (filter.typeProperty?.values) {
        for (const value of Object.values(filter.typeProperty.values)) {
          options.push({ value, label: t(value) });
        }
      }
    } else if (filter.type === "Person") {
      if (filter?.members?.length) {
        for (const member of Object.values(filter.members)) {
          options.push({ value: member.user.name, label: member.user.name });
        }
      }
    } else if (filter.type === "Bool") {
      options.push({ value: "true", label: "True" }, { value: "false", label: "False" });
    }

    return options;
  }, [filter, t]);
```

web/src/components/molecules/Content/Table/DropdownRender.tsx
```diff
@@ -0,0 +1,295 @@
...
+    }
+  }, [open, form, defaultValue]);
+
+  const options: {
+    operatorType: string;
+    value: Operator | SortDirection;
+    label: string;
+  }[] = [];
+
+  if (isFilter) {
```
nourbalaha, line 62: Wrap this inside a use memo with the if and switch statements
```suggestion
  const options = useMemo<{
    operatorType: string;
    value: Operator | SortDirection;
    label: string;
  }>(() => {
    const result = [];
   
    if (isFilter) {

nourbalaha (changes requested):

PR #910, opened 2023-10-29
feat(web): sort and column function in views

coolguy741:
# Overview
added sorting and column function in views


server/internal/infrastructure/mongo/item.go
```diff
@@ -406,6 +406,113 @@ func fieldKey(f view.FieldSelector) string {
...
+	return ff
+}
+
+func fieldKey(f view.FieldSelector) string {
+	if f.Type == view.FieldTypeMetaField || f.Type == view.FieldTypeField {
+		return "__temp." + f.ID.String()
+	}
+	return string(f.Type)
+}
+
```
nourbalaha, line 515: Please revert the server changes as this is a web PR and it is duplicated.

web/src/components/molecules/Content/Table/index.tsx
```diff
@@ -92,6 +96,8 @@ const ContentTable: React.FC<Props> = ({
   pageSize,
   requests,
   addItemToRequestModalShown,
+  columns,
+  setColumns,
```
nourbalaha, line 100: Instead of passing these as props you can move the useState from the hooks to this file since it is only used here.

web/src/gql/queries/item.ts
```diff
@@ -121,43 +121,45 @@ export const GET_ITEMS_BY_IDS = gql`
...
           url
         }
-        createdBy {
-          ... on Integration {
-            name
-          }
-          ... on User {
-            name
-          }
-        }
```
nourbalaha, line 149: Why did you remove createdBy?

web/src/gql/queries/item.ts
```diff
@@ -206,7 +208,7 @@ export const UPDATE_ITEM = gql`
         schemaId
         fields {
           value
-          type
+          types
```
nourbalaha, line 211:
```suggestion
          type
```

web/src/components/molecules/Content/Table/index.tsx
```diff
@@ -509,15 +566,28 @@ const ContentTable: React.FC<Props> = ({
           dataSource={contentTableFields}
           tableAlertOptionRender={AlertOptions}
           rowSelection={rowSelection}
-          columns={[...actionsColumn, ...contentTableColumns]}
+          columns={[...actionsColumn, ...contentColumn]}
```
nourbalaha, line 569: This line is giving an error because the contentColumn might be undefined.
add something like this:
```
const tableColumns = useMemo(() => {
    const items = [...actionsColumn];
    if (contentColumn) items.push(...contentColumn);
    return items;
  }, [actionsColumn, contentColumn]);
  ```

File: web/src/components/molecules/Content/Table/index.tsx [hunk 6]
```diff
@@ -509,15 +566,28 @@ const ContentTable: React.FC<Props> = ({
...
+          columnsState={{
+            value: columns,
+            onChange: setColumns,
+          }}
           onChange={(pagination, _, sorter: any) => {
+            console.log("sorter", sorter);
+            console.log("contentTableFields", contentTableFields);
+            console.log("contentTableColumns", contentTableColumns);
```
nourbalaha, line 577: remove `console.log` statements from PRs.

web/src/components/molecules/Content/Table/index.tsx
```diff
@@ -196,6 +238,21 @@ const ContentTable: React.FC<Props> = ({
     [t, sort?.field?.type, sort?.direction, selectedItem?.id, onItemSelect],
   );
 
+  const contentColumn: ProColumns<ContentTableField>[] | undefined = useMemo(
```
nourbalaha, line 241:
```suggestion
  const contentColumns: ProColumns<ContentTableField>[] | undefined = useMemo(
```

nourbalaha (changes requested):

web/src/components/molecules/Content/Table/index.tsx
```diff
@@ -196,6 +238,21 @@ const ContentTable: React.FC<Props> = ({
...
+    () =>
+      contentTableColumns?.map(column => ({
```
caichi-t, line 243: I think it's better that adding property to contentTableColumns in src\components\organisms\Project\Content\ContentList\hooks.ts, but if you did that, is there any problems?

web/src/components/molecules/Content/Table/index.tsx
coolguy741: It's because it will be used for integrating into views menu

web/src/gql/queries/item.ts
coolguy741: When graphQL API that requests createdBy user quits the local server(http://localhost:8080/) on my local environment. 
It will add createBy and updatedBy when this is fixed.

web/src/gql/queries/item.ts
coolguy741: Is it changed to "types"?

web/src/components/molecules/Content/Table/index.tsx
coolguy741: No problem on that, I changed it because it would affect the filtering.
If filtering is ok, we can use "ProColumns<ContentTableField>[] | undefined".

caichi-t: Filtering is ok! You can add the properties in hooks.ts!

web/src/components/molecules/Content/Table/index.tsx
nourbalaha: OK, so you can keep it the way it is.

web/src/components/molecules/Content/Table/index.tsx
coolguy741: I did this to separate data fields and visualization fields.

nourbalaha (approved):

PR #924, opened 2023-11-02
feat(web): integrated sorting and columns with view

coolguy741:
# Overview
Integrated sorting and columns with views


web/src/components/molecules/Content/Table/index.tsx
```diff
@@ -120,6 +123,7 @@ const ContentTable: React.FC<Props> = ({
   const [searchParams, setSearchParams] = useSearchParams();
   const [currentWorkspace] = useWorkspace();
   const t = useT();
+
   const actionsColumns: ExtendedColumns[] = useMemo(
```
nourbalaha, line 127: Can you please fix the type here, it seems like the types are incompatible.

nourbalaha (changes requested):

nourbalaha (approved):

PR #939, opened 2023-11-08
fix(web): supporting views

coolguy741:
# Overview
refactoring initialization part of views.
fixed navigating newly created view


web/src/components/organisms/Project/Content/ViewsMenu/hooks.ts
```diff
@@ -41,20 +41,16 @@ export default ({ modelId, currentView, setCurrentView }: Params) => {
   });
 
   const views = useMemo(() => {
-    if (!loading && data?.view) {
+    if (loading && data?.view) {
```
nourbalaha, line 44: Why did you remove the exclamation mark? this is causing the initial view to not be loaded correctly.

nourbalaha (changes requested):

web/src/components/organisms/Project/Content/ViewsMenu/hooks.ts
coolguy741: With exclamation mark, navigating to newly created view doesn't work.

nourbalaha: initial view and newly created view should be working at the same time

caichi-t: I think he did it for navigating to newly created view but it cause the issue so we should do other solution.

coolguy741: Sorry. I was wrong. I created prevModelId state.

nourbalaha (approved):

PR #943, opened 2023-11-09
fix(web): view tabs style

caichi-t:
# Overview
Fixed view tabs style according to design.

### DIFF
--- a/web/src/components/molecules/View/viewMenuItem.tsx
+++ b/web/src/components/molecules/View/viewMenuItem.tsx
@@ -71,31 +71,19 @@ const ViewsMenuItem: React.FC<Props> = ({
   ];
 
   return (
-    <StyledDropdownButton
-      trigger={["click"]}
-      type="text"
-      icon={<Icon icon="more" />}
-      menu={{ items: children }}>
+    <Wrapper>
       {t(view.name)}
-    </StyledDropdownButton>
+      <Dropdown trigger={["click"]} menu={{ items: children }}>
+        <Icon icon="more" size={16} style={{ marginRight: 0 }} />
+      </Dropdown>
+    </Wrapper>
   );
 };
 
 export default ViewsMenuItem;
 
-const StyledDropdownButton = styled(Dropdown.Button)`
+const Wrapper = styled.div`
   display: flex;
   align-items: center;
   gap: 8px;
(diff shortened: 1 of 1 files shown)


web/src/components/molecules/View/viewMenuItem.tsx
```diff
@@ -71,31 +71,19 @@ const ViewsMenuItem: React.FC<Props> = ({
   ];
 
   return (
-    <StyledDropdownButton
-      trigger={["click"]}
-      type="text"
-      icon={<Icon icon="more" />}
-      menu={{ items: children }}>
+    <Wrapper>
       {t(view.name)}
```
nourbalaha, line 75: Not related to your PR but can you please fix it? we don't need t() in here.
```suggestion
      {view.name}
```

nourbalaha (changes requested):

web/src/components/molecules/View/viewMenuItem.tsx
caichi-t: Yes, sure!

web/src/components/molecules/View/viewMenuItem.tsx
```diff
@@ -71,31 +71,19 @@ const ViewsMenuItem: React.FC<Props> = ({
...
-      trigger={["click"]}
-      type="text"
-      icon={<Icon icon="more" />}
-      menu={{ items: children }}>
-      {t(view.name)}
-    </StyledDropdownButton>
+    <Wrapper>
+      {view.name}
+      <Dropdown trigger={["click"]} menu={{ items: children }}>
+        <Icon icon="more" size={16} style={{ marginRight: 0 }} />
```
nourbalaha, line 77: Can you use styled components instead of inline styles.
Inline styles are slow and they decrease the performance. Try to avoid it as much as possible.
```suggestion
        <Icon icon="more" size={16} />
```

nourbalaha (changes requested):

web/src/components/molecules/View/viewMenuItem.tsx
caichi-t: Okay!

nourbalaha (approved):

PR #946, opened 2023-11-09
refactor(web): advanced filter's sort code

coolguy741:
# Overview
changed sorter type to ItemSortType
Removed sort URL parameters


web/src/components/molecules/Content/Table/DropdownRender.tsx
```diff
@@ -328,20 +329,31 @@ const DropdownRender: React.FC<Props> = ({
...
+      let fieldType: FieldType;
+      let fieldId = "";
       switch (filter.id as string) {
         case "CREATION_DATE":
         case "CREATION_USER":
         case "MODIFICATION_DATE":
         case "MODIFICATION_USER":
         case "STATUS":
-          searchParams.set("sortFieldType", filter.id);
+          fieldType = filter.id as FieldType;
```
nourbalaha, line 342: This code is confusing? 
Please stop using api types directly, instead create local types and make converter functions to convert from api type to local type.
And try to avoid type assertions as much as possible.
We are using typescript not javascript, so types should be taken in consideration.

nourbalaha (changes requested):

web/src/components/molecules/Content/Table/DropdownRender.tsx
coolguy741: filter.id is the key of a column that the user wants to filter.
If a column is a meta field or field, then the key of the column is corresponding field ID.
If a column represents the data of item's information (ex: createdAt, createdBy), the key of column is the fieldType of the corresponding data(ex: "CREATION_DATE", "CREATION_USER")

web/src/components/organisms/Project/Content/ViewsMenu/hooks.ts
```diff
@@ -42,29 +46,34 @@ export default ({ modelId, currentView, setCurrentView }: Params) => {
...
+    const viewList = data?.view
+      ?.map<View | undefined>(view => fromGraphQLView(view as GQLView))
+      .filter((view): view is View => !!view);
+
+    if (prevModelId !== modelId && viewList) {
+      setSelectedView(viewList && viewList.length > 0 ? viewList[0] : undefined);
       setPrevModelId(modelId);
     }
-    return data?.view ? data?.view : [];
+    return viewList ? viewList : [];
```
nourbalaha, line 57:
```suggestion
    return viewList ?? [];
```

web/src/utils/values.ts
```diff
@@ -94,5 +105,42 @@ export const fromGraphQLView = (view: GQLView): View | undefined => {
     name: view.name,
     modelId: view.modelId,
     projectId: view.projectId,
+    sort: view.sort
+      ? {
+          field: {
+            id: view.sort.field.id ? view.sort.field.id : undefined,
```
nourbalaha, line 111:
```suggestion
            id: view.sort.field.id ?? undefined,
 ```

File: web/src/utils/values.ts [hunk 4]
```diff
@@ -94,5 +105,42 @@ export const fromGraphQLView = (view: GQLView): View | undefined => {
...
+            type: view.sort.field.type,
+          },
+          direction: view.sort.direction ? view.sort.direction : "ASC",
+        }
+      : undefined,
+    columns: view.columns
+      ? view.columns?.map(column => ({
+          type: column.type,
+          id: column.id ? column.id : undefined,
```
nourbalaha, line 120:
```suggestion
          id: column.id ?? undefined,
```

web/src/utils/values.ts
```diff
@@ -94,5 +105,42 @@ export const fromGraphQLView = (view: GQLView): View | undefined => {
...
+        }))
+      : undefined,
+    filter: view.filter ? (view.filter as Condition) : undefined,
+  };
+};
+
+export const toGraphItemSort = (sort: ItemSort): GQLItemSortInput | undefined => {
+  return {
+    field: {
+      id: sort.field.id ? sort.field.id : undefined,
```
nourbalaha, line 130:
```suggestion
      id: sort.field.id ?? undefined,
```

nourbalaha (approved):

PR #947, opened 2023-11-10
chore(server): support group fields in public api

mimoham24:
# Overview

### DIFF
--- a/server/internal/adapter/publicapi/item.go
+++ b/server/internal/adapter/publicapi/item.go
@@ -3,6 +3,9 @@ package publicapi
 import (
 	"context"
 	"errors"
+	"github.com/reearth/reearth-cms/server/internal/adapter"
+	"github.com/reearth/reearth-cms/server/pkg/group"
+	"github.com/reearth/reearth-cms/server/pkg/value"
 
 	"github.com/reearth/reearth-cms/server/pkg/asset"
 	"github.com/reearth/reearth-cms/server/pkg/id"
@@ -58,8 +61,12 @@ func (c *Controller) GetItem(ctx context.Context, prj, mkey, i string) (Item, er
 			return Item{}, err
 		}
 	}
+	sgl, err := getGroupSchemas(ctx, itv, s)
 
-	return NewItem(itv, s, assets, c.assetUrlResolver), nil
+	if err != nil {
+		return Item{}, err
+	}
+	return NewItem(itv, s, sgl, assets, c.assetUrlResolver), nil
 }
 
 func (c *Controller) GetItems(ctx context.Context, prj, model string, p ListParam) (ListResult[Item], *schema.Schema, error) {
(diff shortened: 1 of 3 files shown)


server/internal/adapter/publicapi/types.go
```diff
@@ -55,8 +55,9 @@
 }
 
 type Item struct {
-	ID     string
-	Fields ItemFields
+	ID          string
+	Fields      ItemFields
+	GroupFields ItemFields
```
yk-eukarya, line 60:
```suggestion
```

yk-eukarya: not used

server/internal/adapter/publicapi/types.go
```diff
@@ -118,6 +124,18 @@
...
+			for _, v := range f.Value().Values() {
+				itgID, ok := v.ValueGroup()
+				if !ok {
+					continue
+				}
+				gf := fields.FieldsByGroup(itgID)
+				igf := NewItemFields(gf, groupFields, nil, assets, urlResolver)
+				res[itgID.String()] = igf
+			}
+			val = res
```
yk-eukarya, line 138: multiple group is not handled correctly, check schema field to determine if the field is multiple or not.
output example for single:
```json
...
"groupKey": {
   "groupFieldKey": "groupFieldValue",
   ...
}
```

output example for multiple:
```json
...
"groupKey": [
   {
      "groupFieldKey": "groupFieldValue",
      ...
   },
   {}
]
```

server/internal/adapter/publicapi/types.go
mimoham24: sorry, it was an old implementation.

server/internal/adapter/publicapi/types.go
```diff
@@ -65,10 +65,15 @@
 	return json.Marshal(m)
 }
 
-func NewItem(i *item.Item, s *schema.Schema, assets asset.List, urlResolver asset.URLResolver) Item {
+func NewItem(i *item.Item, s *schema.Schema, groupSchemas schema.List, assets asset.List, urlResolver asset.URLResolver) Item {
```
yk-eukarya, line 68: its better to refactor it to use schema package

server/internal/adapter/publicapi/item.go
```diff
@@ -97,8 +104,50 @@
 		}
 	}
 
-	res := NewListResult(util.Map(items.Unwrap(), func(i *item.Item) Item {
-		return NewItem(i, s, assets, c.assetUrlResolver)
-	}), pi, p.Pagination)
+	itms, err := util.TryMap(items.Unwrap(), func(i *item.Item) (Item, error) {
+		sgl, err := getGroupSchemas(ctx, i, s)
```
yk-eukarya, line 108: querying the db inside a loop will cause performance issues!

server/internal/adapter/publicapi/types.go
mimoham24: multiple fields should show like the following:
```
"groupFiledKey": {
    "itemGroupFieldID1": {
       "field1Key": "value",
       "field2Key": "value",
    },
    "itemGroupFieldID2": {
       "field1Key": "value",
       "field2Key": "value",
    },
    "itemGroupFieldID3": {
       "field1Key": "value",
       "field2Key": "value",
    }
  }
  ```
  I think it's important to show the itemGroup IDs as well, so a map is the better than an array in this case

server/internal/adapter/publicapi/types.go
mimoham24: I know that the function can be prettier, but as the group ID is not involved in the implementation of `NewItem` I think calculating it will just add extra work

yk-eukarya (approved):

PR #949, opened 2023-11-13
fix(web): Unify table controls with setcurrentview

caichi-t:
# Overview
Removed `onTableControl` function and use `setcurrentview` instead.

## Memo
In addition, fixed a small bug related to boolean. (filterDropdown.tsx:306-308)


web/src/components/molecules/Content/Table/index.tsx
```diff
@@ -37,13 +37,7 @@ import {
...
 import { Request } from "@reearth-cms/components/molecules/Request/types";
 import { CurrentViewType } from "@reearth-cms/components/organisms/Project/Content/ContentList/hooks";
-import {
-  SortDirection,
-  FieldSelector,
-  FieldType,
-  ItemSortInput,
-  ConditionInput,
-} from "@reearth-cms/gql/graphql-client-api";
+import { SortDirection, FieldSelector, FieldType } from "@reearth-cms/gql/graphql-client-api";
```
nourbalaha, line 40: GraphQL types shouldn't be used in molecules.
ViewModel types with a converter function to convert from GraphQL type to the ViewModel type should be created.
Please check the reearth frontend development guide for reference.

nourbalaha (changes requested):

nourbalaha (approved):

PR #953, opened 2023-11-15
fix(web): redundant codes and type assertion

caichi-t:
## What I've done
- Fixed the sort operator initialization issue
- Fixed the redundant codes
- Avoided using assertions as much as possible and added processes of type judgments.
## Memo
In some cases, it may be better to review the original type definitions and reduce the processes of type judgments.

### DIFF
--- a/web/src/components/molecules/Content/Table/index.tsx
+++ b/web/src/components/molecules/Content/Table/index.tsx
@@ -640,10 +640,12 @@ const ContentTable: React.FC<Props> = ({
       }
       const cols: FieldSelector[] = tableColumns
         .filter(col => {
-          return typeof col.key === "string" &&
-              col.key !== "EDIT_ICON" &&
-              col.key !== "commentsCount" &&
-              !hiddenCols.includes(col.key)
+          return (
+            typeof col.key === "string" &&
+            col.key !== "EDIT_ICON" &&
+            col.key !== "commentsCount" &&
+            !hiddenCols.includes(col.key)
+          );
         })
         .map(col => {
           if (
--- a/web/src/components/molecules/Content/Table/index.tsx
+++ b/web/src/components/molecules/Content/Table/index.tsx
@@ -19,7 +19,6 @@ import Dropdown, { MenuProps } from "@reearth-cms/components/atoms/Dropdown";
 import Icon from "@reearth-cms/components/atoms/Icon";
 import Input from "@reearth-cms/components/atoms/Input";
 import {
-  ProColumns,
   TableRowSelection,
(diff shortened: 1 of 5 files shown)


web/src/components/molecules/Content/Table/index.tsx
```diff
@@ -258,57 +226,20 @@ const ContentTable: React.FC<Props> = ({
...
-  );
+    ];
+  }, [t, currentView.sort, selectedItem?.id, onItemSelect]);
 
   const tableColumns = useMemo(() => {
-    const items = [...actionsColumns];
-    if (contentColumns) items.push(...contentColumns);
-    return items;
-  }, [actionsColumns, contentColumns]);
+    return contentTableColumns ? [...actionsColumns, ...contentTableColumns] : actionsColumns;
```
nourbalaha, line 241:
```suggestion
    return contentTableColumns ? [...actionsColumns, ...contentTableColumns] : [...actionsColumns];
```

web/src/components/molecules/Content/Table/index.tsx
```diff
@@ -362,37 +293,51 @@ const ContentTable: React.FC<Props> = ({
...
-          if (c.key === fieldId.id) {
-            column = c;
-            break;
+        const column = (() => {
+          for (const c of columns) {
+            if (c.key === fieldId.id) {
+              return c;
+            }
+          }
+        })();
```
nourbalaha, line 314:
```suggestion
        const column = columns.find(c => c.key === fieldId.id);
```

web/src/components/molecules/Content/Table/index.tsx
```diff
@@ -694,33 +646,38 @@ const ContentTable: React.FC<Props> = ({
...
+            if (
+              col.key === "EDIT_ICON" ||
+              col.key === "commentsCount" ||
+              hiddenCols.includes(col.key)
+            ) {
+              return false;
+            } else {
+              return true;
+            }
+          }
```
nourbalaha, line 659:
```suggestion
          return typeof col.key === "string" &&
              col.key !== "EDIT_ICON" &&
              col.key !== "commentsCount" &&
              !hiddenCols.includes(col.key)
```

nourbalaha (changes requested):

web/src/components/molecules/Content/Table/index.tsx
caichi-t: Thank you! I wrote the weird code. lol

nourbalaha (approved):

PR #957, opened 2023-11-21
feat(web): add link to Re:Earth workspace on header

caichi-t:
## What I've done
- added the environment variable of Re:Earth domain


web/src/components/molecules/Common/Header/index.tsx
```diff
@@ -146,6 +146,14 @@ const HeaderMolecule: React.FC<Props> = ({
       )}
       <Spacer />
       <AccountDropdown name={username} items={AccountItems} personal={true} />
+      <LinkWrapper>
+        <EditorLink
+          rel="noopener noreferrer"
+          href={`${window.REEARTH_CONFIG?.webDomain}dashboard/${currentWorkspace?.id}`}
```
nourbalaha, line 152: Just a suggestion, you can ignore it:
`const url = new URL(`/dashboard/${currentWorkspace?.id}`, window.REEARTH_CONFIG?.webDomain ?? "");`
```suggestion
          href={url.toString()}

nourbalaha (approved):

web/src/components/molecules/Common/Header/index.tsx
caichi-t: It seems to crash if the second argument of new URL is "".

nourbalaha: I think you added a default value so it will not be "", any ways you can ignore it. it is just a suggestion.

PR #974, opened 2023-11-30
fix(web): editor URL setting

caichi-t:
# Overview
This PR makes it possible to config `editorUrl` in `reearth_config.json`.
If you set `http(s)://ANY_DOMAIN(/)` to the value, the `Go to Editor` button shows up.

### DIFF
--- a/web/src/components/molecules/Common/Header/index.tsx
+++ b/web/src/components/molecules/Common/Header/index.tsx
@@ -39,10 +39,12 @@ const HeaderMolecule: React.FC<Props> = ({
   const t = useT();
   const { logout } = useAuth();
   const navigate = useNavigate();
-  const url =
-    window.REEARTH_CONFIG?.editorUrl &&
-    currentWorkspace?.id &&
-    new URL(currentWorkspace.id, window.REEARTH_CONFIG?.editorUrl);
+  const url = useMemo(() => {
+    if (window.REEARTH_CONFIG?.editorUrl && currentWorkspace?.id) {
+      return new URL(`dashboard/${currentWorkspace.id}`, window.REEARTH_CONFIG?.editorUrl);
+    }
+    return undefined;
+  }, [currentWorkspace?.id]);
 
   const currentIsPersonal = useMemo(
     () => currentWorkspace?.id === personalWorkspace?.id,
--- a/web/src/components/molecules/Common/Header/index.tsx
+++ b/web/src/components/molecules/Common/Header/index.tsx
@@ -39,6 +39,10 @@ const HeaderMolecule: React.FC<Props> = ({
   const t = useT();
   const { logout } = useAuth();
   const navigate = useNavigate();
+  const url =
+    window.REEARTH_CONFIG?.editorUrl &&
+    currentWorkspace?.id &&
+    new URL(currentWorkspace.id, window.REEARTH_CONFIG?.editorUrl);
 
   const currentIsPersonal = useMemo(
     () => currentWorkspace?.id === personalWorkspace?.id,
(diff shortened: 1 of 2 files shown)


web/src/components/molecules/Common/Header/index.tsx
```diff
@@ -39,6 +39,10 @@ const HeaderMolecule: React.FC<Props> = ({
   const t = useT();
   const { logout } = useAuth();
   const navigate = useNavigate();
+  const url =
+    window.REEARTH_CONFIG?.editorUrl &&
+    currentWorkspace?.id &&
+    new URL(currentWorkspace.id, window.REEARTH_CONFIG?.editorUrl);
```
nourbalaha, line 45:
```suggestion
  const url = useMemo(() => {
    if (window.REEARTH_CONFIG?.editorUrl && currentWorkspace?.id) {
      return new URL(currentWorkspace.id, window.REEARTH_CONFIG?.editorUrl);
    }
    return undefined;
  }, [currentWorkspace?.id]);
```

web/src/config.ts
```diff
@@ -15,21 +15,21 @@ export type Config = {
...
 const env = import.meta.env;
 
 export const defaultConfig: Config = {
   api: env.REEARTH_CMS_API || "/api",
-  auth0Audience: env.REEARTH_CMS_AUTH0_AUDIENCE || "https://api.test.reearth.dev",
-  auth0Domain: env.REEARTH_CMS_AUTH0_DOMAIN || "reearth-oss-test.eu.auth0.com",
-  auth0ClientId: env.REEARTH_CMS_AUTH0_CLIENT_ID || "k6F1sgFikzVkkcW9Cpz7Ztvwq5cBRXlv",
+  auth0Audience: env.REEARTH_CMS_AUTH0_AUDIENCE,
+  auth0Domain: env.REEARTH_CMS_AUTH0_DOMAIN,
+  auth0ClientId: env.REEARTH_CMS_AUTH0_CLIENT_ID,
```
nourbalaha, line 27: Why did you remove the default values?

web/src/config.ts
```diff
@@ -15,21 +15,21 @@ export type Config = {
...
   authProvider: env.REEARTH_CMS_AUTH_PROVIDER || "auth0",
   logoUrl: env.REEARTH_CMS_LOGO_URL,
   coverImageUrl: env.REEARTH_CMS_COVER_URL,
   cesiumIonAccessToken: env.REEARTH_CMS_CESIUM_ION_ACCESS_TOKEN || "",
-  webDomain: env.REEARTH_WEB_DOMAIN || "https://test.reearth.dev/",
+  editorUrl: env.REEARTH_CMS_EDITOR_DOMAIN,
```
nourbalaha, line 32:
```suggestion
  editorUrl: env.REEARTH_CMS_EDITOR_URL,
```

nourbalaha (changes requested):

web/src/config.ts
caichi-t: I discussed about this with Inoue-san, and he said the default values should not be made public since the test environment is not open for general users.

web/src/config.ts
caichi-t: I made a mistake... Thank you!!

web/src/config.ts
nourbalaha: Okay, then you can remove it.

web/src/components/molecules/Common/Header/index.tsx
caichi-t: What do you think about the path (`dashboard/`), which could change depending on the environment?
Should I code it?

web/src/config.ts
nourbalaha: Thank you!

web/src/components/molecules/Common/Header/index.tsx
caichi-t: Okay, I'll fix it. Thank you!

nourbalaha (approved):

PR #981, opened 2023-12-05
feat(server): model ordering

mimoham24:


server/pkg/model/list.go
```diff
@@ -18,3 +19,17 @@ func (l List) SortByID() List {
...
 }
+
+func (l List) OrderByIDs(ids id.ModelIDList) List {
+	var res List
+	for i, mid := range ids {
+		for _, model := range l {
+			if model.ID() == mid {
+				model.SetOrder(i)
+				res = append(res, model)
+				continue
```
yk-eukarya, line 30:
```suggestion
				break
```

server/e2e/gql_model_test.go
```diff
@@ -232,3 +301,55 @@ func TestCreateModel(t *testing.T) {
...
+	res1.
+		Value(2).
+		Object().
+		HasValue("order", 2).
+		HasValue("name", "test2")
+	res1.
+		Value(3).
+		Object().
+		HasValue("order", 3).
+		HasValue("name", "test3")
```
yk-eukarya, line 354:
```suggestion
	res.Path("$.data.updateModesOrder.models[:].id").Array().IsEqual([]string{mId1, mId2})
	res.Path("$.data.updateModesOrder.models[:].order").Array().IsEqual([]int{1, 2})
```

server/internal/infrastructure/memory/model.go
```diff
@@ -179,3 +179,22 @@ func MockModelNow(r repo.Model, t time.Time) func() {
...
+func (r *Model) SaveAll(ctx context.Context, models model.List) error {
+	if r.err != nil {
+		return r.err
+	}
+	if len(models) == 0 {
+		return nil
+	}
+	if !r.f.CanWrite(models[0].Project()) {
+		return repo.ErrOperationDenied
+	}
```
yk-eukarya, line 192: what if not all models on the same project?

server/internal/usecase/interactor/model.go
```diff
@@ -255,3 +255,24 @@ func (i Model) FindOrCreateSchema(ctx context.Context, param interfaces.FindOrCr
...
+			}
+			if !operator.IsMaintainingProject(models[0].Project()) {
+				return nil, interfaces.ErrOperationDenied
+			}
+			ordered := models.OrderByIDs(ids)
+			if err := i.repos.Model.SaveAll(ctx, ordered); err != nil {
+				return nil, err
+			}
+			return ordered, nil
+		})
```
yk-eukarya, line 277:
```suggestion
		func(ctx context.Context) (_ model.List, err error) {
			if len(ids) == 0 {
				return nil, nil
			}
			models, err := i.repos.Model.FindByIDs(ctx, ids)
			if err != nil {
				return nil, err
			}
			if len(models) != len(ids) {
				return nil, // return error here
			}
			if !operator.IsMaintainingProject(models[0].Project()) {
				return nil, interfaces.ErrOperationDenied
			}
			ordered := models.OrderByIDs(ids)
			if err := i.repos.Model.SaveAll(ctx, ordered); err != nil {
				return nil, err
			}
			return ordered, nil
		})
```

server/internal/usecase/interactor/model.go
```diff
@@ -255,3 +255,24 @@ func (i Model) FindOrCreateSchema(ctx context.Context, param interfaces.FindOrCr
...
+	return Run1(ctx, operator, i.repos, Usecase().Transaction(),
+		func(ctx context.Context) (_ model.List, err error) {
+			models, err := i.repos.Model.FindByIDs(ctx, ids)
+			if err != nil {
+				return nil, err
+			}
+			if len(models) == 0 {
+				return nil, nil
+			}
+			if !operator.IsMaintainingProject(models[0].Project()) {
```
yk-eukarya, line 269: what if not all models on the same project?

server/internal/infrastructure/memory/model.go
mimoham24: I was thinking of the same thing, I'm gonna check for that and return an error

yk-eukarya (approved):

PR #1001, opened 2023-12-13
fix(web): the search function on the request modal

caichi-t:
## What I've done
- Fixed  the search function and the styling on the request modal
- Added header to Request page

##Memo
In addition, I fixed the behavior after canceling delete a view.


web/src/components/molecules/Request/List/index.tsx
```diff
@@ -60,6 +61,7 @@ const RequestListMolecule: React.FC<Props> = ({
     <ComplexInnerContents
       center={
         <Content>
+          <StyledPageHeader title="Request" />
```
nourbalaha, line 64:
```suggestion
          <StyledPageHeader title={t("Request")} />
```

nourbalaha (changes requested):

nourbalaha (approved):

PR #1037, opened 2024-01-25
fix(web): update the item with a new group field

caichi-t:
# Overview
This PR fixes the bug that the same value is input in any group field added later and we can't update item after adding a new group field.

## What I've done
- I assigned a new ID for a new group field.
- I refactored some code to fix the bug.
- I fixed the change detection for a single group field.


web/src/components/molecules/Common/Form/AssetItem/index.tsx
```diff
@@ -121,7 +121,7 @@ const AssetItem: React.FC<Props> = ({
       {value ? (
         <>
           <AssetDetailsWrapper>
-            <AssetButton enabled={!!asset} disabled={disabled} onClick={handleClick}>
+            <AssetButton enabled={asset && 1} disabled={disabled} onClick={handleClick}>
```
nourbalaha, line 124: What does `enabled={asset && 1}` do?

web/src/components/molecules/Common/Form/AssetItem/index.tsx
```diff
@@ -131,7 +131,7 @@ const AssetItem: React.FC<Props> = ({
               <Link
                 to={`/workspace/${workspaceId}/project/${projectId}/asset/${value}`}
                 target="_blank">
-                <AssetLinkedName enabled={!!asset} type="link">
+                <AssetLinkedName enabled={asset && 1} type="link">
```
nourbalaha, line 134: Same here, what does `enabled={asset && 1}` do?

nourbalaha (approved):

PR #1039, opened 2024-01-29
fix(web): minor bugs in appearance, etc.

caichi-t:
# Overview
This PR includes to fix the default terrain.


web/src/components/molecules/Common/Form/GroupItem/index.tsx
```diff
@@ -121,24 +121,36 @@ const GroupItem: React.FC<Props> = ({
...
+                }}>
+                <Icon icon="arrowDown" />
+              </IconWrapper>
+              <IconWrapper
+                onClick={e => {
+                  onDelete?.();
+                  e.stopPropagation();
+                }}>
+                <Icon icon="delete" />
+              </IconWrapper>
```
nourbalaha, line 153: Just a suggestion, it would be better to move the onClick functions from here and create functions outside the return block.
```suggestion
              <IconWrapper
                disabled={disableMoveUp}
                onClick={handleMoveUp}}>
                <Icon icon="arrowUp" />
              </IconWrapper>
              <IconWrapper
                disabled={disableMoveDown}
                onClick={handleMoveDown}>
                <Icon icon="arrowDown" />
              </IconWrapper>
              <IconWrapper
                onClick={handleDelete}>
                <Icon icon="delete" />
              </IconWrapper>
```

nourbalaha (approved):

web/src/components/molecules/Common/Form/GroupItem/index.tsx
caichi-t: Thank you, I think so, too. I'll fix it.

PR #1070, opened 2024-02-19
fix(web): asset value bugs

caichi-t:
# Overview
This PR fixes the bug about the asset field.
I decided to use getAsset query to get default asset name or new asset name that the item doesn't have.

## What I've done
- Fixed the bug that the default asset value is displayed as a removed asset when creating a new item.
- Fixed the bug that the changing group order make asset name as a removed asset
- Fixed the bug that we can’t select the same asset once unlinking


web/src/components/molecules/Common/Form/AssetItem/index.tsx
```diff
@@ -96,19 +98,37 @@ const AssetItem: React.FC<Props> = ({
...
+      setAsset(assetInfo);
+    } else if (selectedAssetRef.current) {
       setAsset(selectedAssetRef.current);
+    } else if (value) {
+      defaultValueGet(value);
+    } else {
+      setAsset(undefined);
     }
-  }, [itemAssets, value]);
+  }, [defaultValueGet, value, loadingAssets, itemAssets]);
```
nourbalaha, line 121: Can you rethink this login?

nourbalaha (changes requested):

web/src/components/molecules/Common/Form/AssetItem/index.tsx
caichi-t: I think this is the only way to avoid throwing getAsset query in vain.
If the result of the getItem query could include a default asset name, it would be simple a bit.

nourbalaha (approved):

PR #1073, opened 2024-02-21
fix(server): cannot unset a schema field title

mimoham24:


server/pkg/schema/schema.go
```diff
@@ -103,6 +103,12 @@ func (s *Schema) TitleField() *FieldID {
...
 func (s *Schema) SetTitleField(tf *FieldID) error {
+	// unsetting title
+	if tf == nil {
+		s.titleField = nil
+		return nil
+	}
+
 	if !s.HasField(*tf) || s.Fields() == nil || len(s.Fields()) == 0 {
 		s.titleField = nil
 		return ErrInvalidTitleField
```
yk-eukarya, line 114:
```suggestion
	if !s.HasField(*tf) || s.Fields() == nil || len(s.Fields()) == 0 {
		return ErrInvalidTitleField
```

yk-eukarya (commented):
we should also handle the case of deleting a field set to `titleField`

yk-eukarya (approved):

PR #1094, opened 2024-03-10
feat(server,worker): add authentication for AWS SNS subscriptions

int-tt:
# Overview
Add support because AWS SNS has the following differences from GCP.
* Authentication required when building SNS subscription
* SNS https endpoint does not support jwt authentication

### DIFF
--- a/server/internal/app/m2m.go
+++ b/server/internal/app/m2m.go
@@ -20,6 +20,13 @@ func NotifyHandler() echo.HandlerFunc {
 		var input rhttp.NotifyInput
 		var err error
 
+		if isAWSSNSSubscriptionConfirmation(c.Request()) {
+			// Handle AWS SNS subscription confirmation
+			// This is used to handle requests for AWS SNS subscription confirmation and is only executed during initial setup
+			// https://docs.aws.amazon.com/sns/latest/dg/SendMessageToHttp.prepare.html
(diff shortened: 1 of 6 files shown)


server/internal/app/usecase.go
```diff
@@ -43,3 +45,14 @@ func ContextMiddleware(fn func(ctx context.Context) context.Context) echo.Middle
 		}
 	}
 }
+
+func AuthTokenMiddleware(token string, key any) echo.MiddlewareFunc {
```
int-tt, line 49: m2mの方に移動する

server/internal/app/m2m.go
```diff
@@ -20,6 +20,13 @@ func NotifyHandler() echo.HandlerFunc {
 		var input rhttp.NotifyInput
 		var err error
 
+		if isAWSSNSSubscriptionConfirmation(c.Request()) {
+			// Handle AWS SNS subscription confirmation
+			// This is used to handle requests for AWS SNS subscription confirmation and is only executed during initial setup
+			// https://docs.aws.amazon.com/sns/latest/dg/SendMessageToHttp.prepare.html
+			return handleSubscriptionConfirmation(c)
+		}
+
```
int-tt, line 29: TODO: workerと同じようにmiddlewareとして切り出す

rot1024 (approved):

PR #1129, opened 2024-04-17
feat(web): show notification when updating member’s role

caichi-t:


web/src/components/organisms/Settings/Members/hooks.ts
```diff
@@ -151,13 +151,20 @@ export default () => {
...
+        if (result.errors || !workspace) {
+          Notification.error({ message: t("Failed to update member's role.") });
+          return;
+        }
         if (workspace) {
           setWorkspace(fromGraphQLWorkspace(workspace as GQLWorkspace));
         }
+        if (result.data) {
+          Notification.success({ message: t("Successfully udpated member's role!") });
+        }
```
nourbalaha, line 164: can you rethink the if statements logic? something like this?

```suggestion
        const workspace = result.data?.updateUserOfWorkspace?.workspace;
        if (result.errors || !workspace) {
          Notification.error({ message: t("Failed to update member's role.") });
          return;
        }
        setWorkspace(fromGraphQLWorkspace(workspace as GQLWorkspace));
        Notification.success({ message: t("Successfully udpated member's role!") });
```

nourbalaha (changes requested):

web/src/components/organisms/Settings/Members/hooks.ts
caichi-t: Okay, I should have referred to `handleMemberRemoveFromWorkspace`, not `handleUsersAddToWorkspace`.
Thank you.

nourbalaha: Thank you!

nourbalaha (approved):

PR #1158, opened 2024-05-21
feat(web): add support for multi-group ordering

caichi-t:


web/src/gql/queries/group.ts
```diff
@@ -120,3 +121,15 @@ export const MODELS_BY_GROUP = gql`
...
 `;
+
+export const UPDATE_GROUPS_ORDER = gql`
+  mutation UpdateGroupsOrder($groupIds: [ID!]!) {
+    updateGroupsOrder(input: { groupIds: $groupIds }) {
+      groups {
+        ... on Group {
+          id
+        }
+      }
```
nourbalaha, line 132: I think there is no need to add `... on Group` here
```suggestion
      groups {
        id
      }
```

nourbalaha (approved):

PR #1163, opened 2024-05-22
feat(web): add support for multi-view ordering

caichi-t:
This PR adds the feature to reorder multi-views.


web/src/components/molecules/View/viewsMenu.tsx
```diff
@@ -14,6 +18,7 @@ interface Props {
   currentView: CurrentView;
   onViewCreateModalOpen: () => void;
   onViewSelect: (key: string) => void;
+  onUpdateViewsOrder: (viewIds: string[]) => void;
```
nourbalaha, line 21: It might be better to write it this way.
```suggestion
  onUpdateViewsOrder: (viewIds: string[]) => Promise<void>;
```

nourbalaha (approved):

PR #1166, opened 2024-05-30
fix(server): sort models, groups and views by order

nourbalaha:
# Overview
This PR returns sorted models, groups and views when querying them


server/internal/usecase/interactor/view.go
```diff
@@ -33,7 +33,12 @@ func (i View) FindByIDs(ctx context.Context, IDs view.IDList, _ *usecase.Operato
...
 
 func (i View) FindByModel(ctx context.Context, mID view.ModelID, _ *usecase.Operator) (view.List, error) {
-	return i.repos.View.FindByModel(ctx, mID)
+	v, err := i.repos.View.FindByModel(ctx, mID)
+	if err != nil {
+		return nil, err
+	}
+	ordered := v.Ordered()
+	return ordered, nil
 }
```
yk-eukarya, line 42:
```suggestion
	return v.Ordered(), nil
}
```

yk-eukarya (approved):

PR #1169, opened 2024-06-03
feat(web): add csv preview support on asset page

caichi-t:
# Overview
This PR adds supporting to preview CSV on Asset page. For now, `lat` and `lng` properties are used as coordinates.


web/src/components/molecules/Asset/Viewers/CsvViewer/Imagery.tsx
```diff
@@ -0,0 +1,80 @@
...
+              image: mapPin,
+              width: 30,
+              height: 30,
+              disableDepthTestDistance: Number.POSITIVE_INFINITY,
+            },
+            properties: obj,
+          });
+        } else {
+          break;
+        }
```
nourbalaha, line 61: There is no need for the else statement here.

web/src/components/molecules/Asset/Viewers/CsvViewer/Imagery.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const lines = text.split(/\r\n|\n|\r/);
+    const headers = lines[0].split(",");
+    for (let i = 1; i < lines.length; i++) {
+      const obj: GeoObj = {};
+      const line = lines[i].split(",");
+      for (let j = 0; j < headers.length; j++) {
+        obj[headers[j]] = line[j];
+      }
+      result.push(obj);
+    }
```
nourbalaha, line 41: Just a suggestion:
If you can use forEach or for ... of statements instead of a normal for loop, it would be better for code readability.

nourbalaha (approved):

nourbalaha (approved):

PR #1186, opened 2024-07-01
fix(web): show loading for various actions

caichi-t:
# Overview
This PR fixes to show loading for various actions.

## Memo
This includes a little styling and translation fixes.


web/src/components/molecules/Content/Table/types.ts
```diff
@@ -1,3 +1,5 @@
+import { t } from "i18next";
```
nourbalaha, line 1: Please add `i18next` to `src/i18n` and then import it from there

web/src/components/molecules/Member/MemberAddModal/index.tsx
```diff
@@ -66,28 +68,18 @@ const MemberAddModal: React.FC<Props> = ({
...
+    await onSubmit(
+      searchedUserList.map(user => ({
+        userId: user.id,
+        role: "READER",
+      })),
+    );
+    changeSearchedUser(undefined);
+    changeSearchedUserList([]);
+    onClose();
+    form.resetFields();
```
nourbalaha, line 82: try, catch... statement should be added here

web/src/components/molecules/Member/MemberRoleModal/index.tsx
```diff
@@ -14,41 +15,55 @@ interface FormValues {
...
-        await onSubmit(member.userId, values.role);
-        onClose();
-        form.resetFields();
-      })
-      .catch(info => {
-        console.log("Validate Failed:", info);
-      });
+  const handleSubmit = useCallback(async () => {
+    const values = await form.validateFields();
+    await onSubmit(member.userId, values.role);
```
nourbalaha, line 37: try, catch... statement should be added here

nourbalaha (changes requested):

web/src/components/molecules/Member/MemberAddModal/index.tsx
caichi-t: There is no process to do within the scope of catch, but is it necessary to use try-catch?
Or is there error handling that should be done?

nourbalaha: I think for now it is okay to do something like this:
```
try {
  // some code
} catch(error) {
  console.error(error);
}
```
maybe later we can work on adding/displaying error messages

nourbalaha: Also these should go in the `finally` block:
```
    changeSearchedUser(undefined);
    changeSearchedUserList([]);
    onClose();
    form.resetFields();
```

caichi-t: > maybe later we can work on adding/displaying error messages

I got it. Is this code better?
```
await onSubmit(
      searchedUserList.map(user => ({
        userId: user.id,
        role: "READER",
      })),
    ).catch(error => {
      console.error(error);
    });
```

> Also these should go in the finally block:

I don't think we can go them in the `finally` block because we want to process them only when onSubmit succeeds.

nourbalaha: Yep, you are right. No need to move them to the finally block.

nourbalaha: > await onSubmit(
>       searchedUserList.map(user => ({
>         userId: user.id,
>         role: "READER",
>       })),
>     ).catch(error => {
>       console.error(error);
>     });

I think since you are using async... await, the try... catch statement is more appropriate. right?

caichi-t: It would be better to use try...catch statement. Thank you for your advice. 🙇‍♂️

nourbalaha: Thank you 🙇🏻

nourbalaha (approved):

PR #1188, opened 2024-07-03
fix(server): geo field missing validations

mimoham24:
# Overview
- geoJSON validation was not following RFC 7946 
- the supported type was not validated. A point can be added to a field that only supports the LineString type without any issues.
- remove GeoJSON feature 
- position validation
- test cases
- support any


server/pkg/schema/field_geometry.go
```diff
@@ -45,40 +55,106 @@ func (f *FieldGeometry) Clone() *FieldGeometry {
...
 		Geometry: func(a value.String) {
-			if !IsValidGeoJSON(a) {
+			t, ok := isValidGeoJSON(a)
+			if !ok {
 				err = ErrInvalidValue
 			}
+			ok2 := f.SupportedTypes().Has(GeometrySupportedTypeFrom(string(t)))
+			if !ok2 {
+				err = ErrUnsupportedType
+			}
```
yk-eukarya, line 157: we have `any` value for the supported types now, so this logic need to be updated

server/pkg/schema/field_geometry_editor.go
```diff
@@ -1,17 +1,20 @@
...
 
 type GeometryEditorSupportedTypeList []GeometryEditorSupportedType
 
+func (l GeometryEditorSupportedTypeList) Has(st GeometryEditorSupportedType) bool {
+	hasAny := slices.Contains(l, GeometryEditorSupportedTypeAny)
+	return slices.ContainsFunc(l, func(t GeometryEditorSupportedType) bool {
+		return t == st || (hasAny && st != "")
+	})
+}
+
```
yk-eukarya, line 17:
```suggestion
func (l GeometryEditorSupportedTypeList) Has(st GeometryEditorSupportedType) bool {
	hasAny := slices.Contains(l, GeometryEditorSupportedTypeAny)
	if hasAny && st != "" {
	     return true
	}
	return slices.ContainsFunc(l, func(t GeometryEditorSupportedType) bool {
		return t == st
	})
}

```

yk-eukarya: also the value should not be `any` 🤔

server/pkg/schema/field_geometry_validation.go
```diff
@@ -0,0 +1,121 @@
...
+//			if !ok2 {
+//				err = ErrUnsupportedType
+//			}
+//		},
+//		Default: func() {
+//			err = ErrInvalidValue
+//		},
+//	})
+//	return
+//}
```
yk-eukarya, line 121:
```suggestion
```

server/pkg/schema/field_geometry_validation.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func isValidRFC7946(g *geojson.Geometry) bool {
+	if g == nil {
+		return false
+	}
+	b := true
+	switch g.Type {
+	case "Point", "MultiPoint":
+		return isValidPosition(g.Point)
```
yk-eukarya, line 67: do the `point` and `multiPoint` share the same validation?

yk-eukarya (approved):

PR #1198, opened 2024-07-30
feat(server): export items as geojson and csv via public api

nourbalaha:
# Overview
This PR implements exporting items as GeoJSON and CSV formats via public api

# What I've done
- export items as GeoJSON via public api
- export items as CSV via public api


server/pkg/exporters/csv.go
```diff
@@ -98,19 +98,25 @@ func extractFirstPointField(itm *item.Item) ([]float64, error) {
 	return nil, noPointFieldError
 }
 
-func convertToCSV(data [][]string) (string, error) {
-	var sb strings.Builder
-	w := csv.NewWriter(&sb)
-	for _, row := range data {
-		if err := w.Write(row); err != nil {
-			return "", err
+func convertToCSV(pw *io.PipeWriter, data [][]string) error {
```
yk-eukarya, line 101: no need to return an error here, instead the error must be checked in from the pipe in the upper layers

yk-eukarya (approved):

PR #1210, opened 2024-08-22
fix(web): upgrade @reearth/core

caichi-t:
# Overview
This PR upgrades @reearth/core to fix the bug that the point is sometimes not rendered in Geometry field.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -135,5 +135,9 @@
     "remark-gfm": "4.0.0",
     "resium": "1.17.4",
     "ulid": "2.3.0"
+  },
+  "resolutions": {
+    "string-width": "4.2.3",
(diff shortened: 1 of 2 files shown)


web/package.json
```diff
@@ -135,5 +135,9 @@
     "remark-gfm": "4.0.0",
     "resium": "1.17.4",
     "ulid": "2.3.0"
+  },
+  "resolutions": {
+    "string-width": "4.2.3",
+    "strip-ansi": "^6.0.1"
```
nourbalaha, line 141: Try this and check if the ci passes
```suggestion
    "strip-ansi": "6.0.1"
```

nourbalaha (changes requested):

nourbalaha (approved):

PR #1225, opened 2024-09-10
fix(web): correct validation behavior when handling invalid values

caichi-t:
# Overview
This PR fixes the behavior when setting validation and some invalid value is present on Text, TextArea, Markdown text, Int, and Geometry fields.

## Memo
This PR includes a small translation fix.


web/src/components/atoms/Input/index.tsx
```diff
@@ -1,6 +1,33 @@
...
+import { useMemo } from "react";
+import { runes } from "runes2";
 
 export type { SearchProps } from "antd/lib/input";
 
+type Props = {
+  value?: string;
+} & InputProps;
+
+const Input: React.FC<Props> = ({ value, maxLength, ...props }, ref) => {
```
nourbalaha, line 11: I think it is better to use `forwardRef` here, you can import `InputRef` from `antd`.
Please do the same for all the other components that pass the ref from a parent component to a child component.
```suggestion
const Input = forwardRef<InputRef, Props>(({ value, maxLength, ...props }, ref) => {

```

web/src/components/atoms/Input/index.tsx
```diff
@@ -1,6 +1,33 @@
...
+  const status = useMemo(() => {
+    if (maxLength && value && runes(value).length > maxLength) {
+      return "error";
+    }
+  }, [maxLength, value]);
+
+  return (
+    <AntInput
```
nourbalaha, line 19: It is better to rename this to a more appropriate name.
Please do the same for all the other components.
```suggestion
    <AntDInput
```

web/src/components/molecules/Common/Form/utils.ts
```diff
@@ -23,3 +23,12 @@ export const keyReplace = (
   const { form, key } = formData;
   form.setFieldValue(key, e.currentTarget.value.replaceAll(" ", "-"));
 };
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const emptyConvert = (value: any) => {
```
nourbalaha, line 28: Please use `unknown` instead of `any`
```suggestion
export const emptyConvert = (value: unknown) => {
```

web/src/components/molecules/Content/Form/fields/FieldComponents/IntegerField.tsx
```diff
@@ -14,6 +16,26 @@ type DefaultFieldProps = {
...
+          if (value < min) {
+            return true;
+          }
+        }
+        if (max) {
+          if (max < value) {
+            return true;
+          }
+        }
+      }
```
nourbalaha, line 34:
```suggestion
    if ((min && value < min) || (max && value > max)) {
      return true;
    }
```

web/src/components/molecules/Content/Form/index.tsx
```diff
@@ -212,18 +213,23 @@ const ContentForm: React.FC<Props> = ({
...
-      return undefined;
-    } else {
-      return value;
-    }
-  }, []);
-
   const handleValuesChange = useCallback(
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    (changedValues: any) => {
+    async (changedValues: any) => {
```
nourbalaha, line 218: Maybe?
```suggestion
    async (changedValues: Record<string, unknown>) => {
```

web/src/components/molecules/Schema/FieldModal/FieldDefaultInputs/IntegerField/index.tsx
```diff
@@ -7,17 +7,53 @@ import { useT } from "@reearth-cms/i18n";
...
+          if (value < min) {
+            return true;
+          }
+        }
+        if (max) {
+          if (max < value) {
+            return true;
+          }
+        }
+      }
```
nourbalaha, line 30:
```suggestion
    if ((min && value < min) || (max && value > max)) {
      return true;
    }
```

nourbalaha (changes requested):

PR #1229, opened 2024-09-12
fix(server): multiple default value bug

nourbalaha:
# Overview
This PR fixes the bug where if there is a field that is multiple and has multiple default values, if the first value is empty all other values are deleted.

### DIFF
--- a/server/internal/adapter/gql/gqlmodel/convert_schema.go
+++ b/server/internal/adapter/gql/gqlmodel/convert_schema.go
@@ -636,9 +636,17 @@ func unpackArray(s any) []any {
 		return nil
 	}
 	v := reflect.ValueOf(s)
+	if v.Kind() != reflect.Slice {
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


server/internal/adapter/gql/gqlmodel/convert_schema.go
```diff
@@ -636,9 +636,17 @@ func unpackArray(s any) []any {
...
 	}
 	v := reflect.ValueOf(s)
+	if v.Kind() != reflect.Slice {
+		return nil
+	}
 	r := make([]any, v.Len())
 	for i := 0; i < v.Len(); i++ {
-		r[i] = v.Index(i).Interface()
+		elem := v.Index(i).Interface()
+		if str, ok := elem.(string); ok && str == "" {
```
yk-eukarya, line 645: It is better to trim the string

yk-eukarya (approved):

PR #1277, opened 2024-10-22
feat(server): implement public api for assets

nourbalaha:
# Overview
This PR introduces the implementation of a public API for assets. 

# What I have done
- Introduced a new endpoint for retrieving project assets: `GET /api/p/{project}/assets`.
- Added methods for batch retrieval of asset files by their IDs.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new endpoint for retrieving project assets: `GET /api/p/{project}/assets`.
	- Enhanced asset retrieval with pagination and detailed JSON responses.
	- Added methods for batch retrieval of asset files by their IDs.
	- Implemented a method to collect asset IDs from a list.

- **Bug Fixes**
	- Modified existing asset retrieval endpoint to include a new `contentType` field for improved data accuracy.
	- Updated test cases to ensure sensitive data is not exposed when projects are private.

- **Refactor**
	- Improved parameter handling and response management in API functions for better clarity and maintainability.


server/internal/adapter/publicapi/asset.go
```diff
@@ -34,3 +38,36 @@ func (c *Controller) GetAsset(ctx context.Context, prj, i string) (Asset, error)
...
+	}, nil)
+
+	if err != nil {
+		if errors.Is(err, rerror.ErrNotFound) {
+			return ListResult[Asset]{}, rerror.ErrNotFound
+		}
+		return ListResult[Asset]{}, err
+	}
+
+	fileMap, err := c.usecases.Asset.FindFilesByIDs(ctx, util.Map(al, func(a *asset.Asset) id.AssetID { return a.ID() }), nil)
```
yk-eukarya, line 65:
```suggestion
	fileMap, err := c.usecases.Asset.FindFilesByIDs(ctx, al.Ids(), nil)
```
implement Ids() method in assetList type

server/internal/usecase/interfaces/asset.go
```diff
@@ -61,6 +61,7 @@ type Asset interface {
 	FindByIDs(context.Context, []id.AssetID, *usecase.Operator) (asset.List, error)
 	FindByProject(context.Context, id.ProjectID, AssetFilter, *usecase.Operator) (asset.List, *usecasex.PageInfo, error)
 	FindFileByID(context.Context, id.AssetID, *usecase.Operator) (*asset.File, error)
+	FindFilesByIDs(context.Context, id.AssetIDList, *usecase.Operator) (map[id.AssetID]*asset.File, error)
```
yk-eukarya, line 64:
```suggestion
	FindFilesByIDs(context.Context, id.AssetIDList, *usecase.Operator) (asset.FileList, error)
```
return list, and implement a method in the list to return a map

nourbalaha: Since `asset.File` does not have assetId it will be impossible to tell which file is for which asset when returning `asset.FileList` that is why I am returning a map with asset id as a key.

```
type File struct {
	name        string
	size        uint64
	contentType string
	path        string
	children    []*File
	files       []*File
}
```

yk-eukarya: the asset file document (which is stored in DB) contains the asset ID, we can expose it to the model 🤔

nourbalaha: Can you explain why is it not a good idea to use a map?

yk-eukarya (approved):

PR #1348, opened 2025-01-14
chore(server): enable to set db name from env file

soneda-yuya:
# Overview
We can't set database name from env file.

## What I've done
- add new env value `DB_CMS`
- read `DB_CMS` value from .env file

## How I tested
- print log on my console which database use

## Summary by CodeRabbit

- **Configuration Updates**
	- Modified environment variables for server and database settings
	- Updated server host configuration
	- Added new database connection parameters

- **Database Initialization**
	- Improved database connection logic
	- Separated account and CMS database initialization
	- Enhanced modularity of database configuration

### DIFF
--- a/server/.env.example
+++ b/server/.env.example
@@ -2,14 +2,15 @@
 PORT=8080
 REEARTH_CMS_DB=mongodb://localhost
 REEARTH_CMS_HOST=https://localhost:8080
-REEARTH_CMS_SERVERHOST=https://localhost:8080
+REEARTH_CMS_SERVERHOST=localhost
 REEARTH_CMS_HOST_WEB=https://localhost:3000
 REEARTH_CMS_ASSETBASEURL=https://localhost:8080/assets
 REEARTH_CMS_DEV=false
 REEARTH_CMS_SIGNUPSECRET=
 REEARTH_CMS_ORIGINS=www.example.com,localhost:3000
 
-REEARTH_CMS_DB_ACCOUNT=mongodb://localhost
+REEARTH_CMS_DB_ACCOUNT=reearth_account
+REEARTH_CMS_DB_CMS=reearth_cms
 REEARTH_CMS_DB_USERS=Test1=mongodb://localhost,Test2=mongodb://localhost
 
 #GraphQL
--- a/server/internal/app/config.go
+++ b/server/internal/app/config.go
@@ -57,8 +57,8 @@ type Config struct {
 	// auth for m2m
 	AuthM2M AuthM2MConfig `pp:",omitempty"`
 
-	DB_Account string          `pp:",omitempty"`
-	DB_CMS     string          `pp:",omitempty"`
+	DB_Account string          `default:"reearth_account" pp:",omitempty"`
+	DB_CMS     string          `default:"reearth_cms" pp:",omitempty"`
 	DB_Users   []appx.NamedURI `pp:",omitempty"`
 }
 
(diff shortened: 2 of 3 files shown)


server/.env.example
```diff
@@ -2,14 +2,15 @@
 PORT=8080
 REEARTH_CMS_DB=mongodb://localhost
 REEARTH_CMS_HOST=https://localhost:8080
-REEARTH_CMS_SERVERHOST=https://localhost:8080
+REEARTH_CMS_SERVERHOST=localhost
```
soneda-yuya, line 5: [memo]
It should be set host but currently set url.

server/internal/app/config.go
```diff
@@ -58,6 +58,7 @@ type Config struct {
 	AuthM2M AuthM2MConfig `pp:",omitempty"`
 
 	DB_Account string          `pp:",omitempty"`
+	DB_CMS     string          `pp:",omitempty"`
```
yk-eukarya, line 61:
```suggestion
	DB_Account string          `default:"reearth_account" pp:",omitempty"`
	DB_CMS     string          `default:"reearth_cms"  pp:",omitempty"`
```

server/internal/app/repo.go
```diff
@@ -23,30 +23,17 @@ import (
...
-		SetMonitor(otelmongo.NewMonitor())
-	if conf.Dev {
-		co.SetMonitor(NewLogMonitor())
-	}
-	client, err := mongo.Connect(ctx, co)
-	if err != nil {
-		log.Fatalf("repo initialization error: %+v\n", err)
-	}
+const defaultCMSDatabase = "reearth_cms"
+const defaultAccountDatabase = "reearth_account"
```
yk-eukarya, line 27:
```suggestion
```
please use config defaults

soneda-yuya: @yk-eukarya 
Thanks!
I fixed it and confirmed.
スクリーンショット 2025-01-14 17 50 52

server/internal/app/repo.go
```diff
@@ -42,31 +73,11 @@ func initReposAndGateways(ctx context.Context, conf *Config) (*repo.Container, *
...
-	}
-
 	txAvailable := mongox.IsTransactionAvailable(conf.DB)
 
-	acRepos, err := accountmongo.New(ctx, client, accountDatabase, txAvailable, false, accountUsers)
-	if err != nil {
-		log.Fatalf("Failed to init mongo: %+v\n", err)
-	}
+	acRepos := initAccountDB(client, txAvailable, ctx, conf)
+	cmsRepos := initCMSDB(client, txAvailable, acRepos, ctx, conf)
```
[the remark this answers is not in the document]
soneda-yuya: In `initXXXXDB`, if there is a problem, the process is stopped using log.Fatal.
Since this is an initialisation process, I do not think it is a problem to process to stop the server instead of returning an error.

server/internal/app/repo.go
```diff
@@ -23,7 +23,38 @@
...
+	cmsDatabase := conf.DB_CMS
+	log.Infof("cms database: %s", cmsDatabase)
+
+	repos, err := mongorepo.New(ctx, client, cmsDatabase, txAvailable, acRepos)
+	if err != nil {
+		log.Fatalf("Failed to init mongo: %+v\n", err)
+	}
+
+	return repos
+}
```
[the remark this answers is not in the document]
soneda-yuya: if `config.DB_CMS` is invalid value, this initialization is failure and stop server in initialization process.
And acRepos and client is be never null, because before executing this function, if these values is null server is stopped by `log.Fatal`
And also this is initialization process, so I think we do not return error.

PR #1404, opened 2025-02-12
test(web): QA examan

patr-coder:
# Overview
Improved our testing framework by making it faster and more maintainable.

## What I've done
* Set up GraphQL API to pre-load test data  (Note: This functionality is currently commented out due to a configuration issue and will be re-enabled once resolved.)

* Switched to data-testid instead of CSS/tags for better element selection
* Split test files for better organization

## What I haven't done
* GraphQL mock data setup still needs configuration fix before enabling

## How I tested
* Run full test suite
* Manual testing of new scripts

* GraphQL API mock data approach
* Usage of data-testid attributes

## Memo
Will enable mock data once config is fixed. Open to feedback on improving parallel execution.


caichi-t:
# Overview
I feel like you just only add comments, adding testid to every element you use for playwright, or just wrapping a function that was already there.

## Need to be fixed
- You shouldn't contain useless files, useless comments, and confidential data.
- There is no need to put together tests that we want to be independent with the utils function. `web/e2e/general/workspace.spec.ts`
- You're using testid for every element, which made it less readable. I feel like you're not using testid where you should, but using it where you don't need to.
- Why you deleted the test code in `web/e2e/settings/settings.spec.ts`?
- The way data-testid is managed is a mystery, some parts are importing and some are writing.
- You shouldn't change the UI by adding a placeholder. `web/src/components/molecules/ProjectSettings/GeneralForm.tsx`
- You shouldn't ignore the warning by ESLint. `web/e2e/project/accessibility.spec.ts`
- You should format the codes.

PR #1477, opened 2025-04-15
ci: add golang license checker

joshuagibeonm:
# Overview

## Summary by CodeRabbit

- **Chores**
  - Added a new automated workflow to check licenses in the server directory during pull requests and manual runs.


yk-eukarya:
@joshuagibeonm I think we should apply this to the worker as well. Also, is it better to move it to the lint workflow `workflows/ci-server.yml`?

PR #1515, opened 2025-06-02
feat(worker): Add Health Check on Worker

jasonkarel:
# Overview
Add health check to worker service

## What I've done
Add health check end-point to worker service

## What I haven't done
Local test of the API running `curl -v http://localhost:8080/health`


worker/internal/app/health.go
```diff
@@ -0,0 +1,81 @@
...
+	"time"
+
+	"cloud.google.com/go/storage"
+	"github.com/hellofresh/health-go/v5"
+	"github.com/hellofresh/health-go/v5/checks/mongo"
+	"github.com/labstack/echo/v4"
+	"github.com/reearth/reearthx/log"
+)
+
+// HealthCheck returns an echo.HandlerFunc that serves the health check endpoint
```
pyshx, line 16: The function name is self explanatory, we don't need these comments.

pyshx (approved):

worker/internal/app/health.go
jasonkarel: deleted

yk-eukarya (approved):

PR #1518, opened 2025-06-04
feat(server): Infrastructure validation in server startup

jasonkarel:
# Overview
This PR introduces several validation checks during server startup to improve system reliability and surface misconfigurations early
## What I've done
✅ Task Runner Validation
Ensure all registered TaskRunner implementations include a HealthCheck method:

```
type TaskRunner interface {
    Run(context.Context, task.Payload) error
    Retry(context.Context, string) error
    HealthCheck(context.Context) error
}
```

✅ Pub/Sub Validation
Validate the Pub/Sub system during startup:

Confirm that Pub/Sub is responding as expected.

✅ Worker Service Validation

✅ Service Account Permissions Check
Validate that the service account used by the server:

Has permission to write logs.

Can access secrets (required for features like item import and copy).

✅ GCP Bucket Validation
Investigate programmatic access to validate:

Bucket accessibility (e.g., whether it's publicly accessible).

GCS client initialization does not return errors.

✅ MongoDB
No additional checks needed — MongoDB connectivity and validation are already handled by existing logic.


server/server
yk-eukarya, line 1: The binary file should not be pushed to the repo!

yk-eukarya (commented):
LGTM

PR #1520, opened 2025-06-05
chore(server): Separate origins for the integration and public api

jasonkarel:
# Overview
Separate origins for the integration and public api
## What I've done
1. Introduce new CORS configuration for integration and public API each
## How I tested
1. tested with different and same origin cases


server/internal/app/app.go
```diff
@@ -108,6 +123,28 @@ func allowedOrigins(appCtx *ApplicationContext) []string {
...
+func allowedPublicOrigins(appCtx *ApplicationContext) []string {
+	if appCtx == nil {
+		return nil
+	}
+	origins := append([]string{}, appCtx.Config.PublicOrigins...)
+	if appCtx.Debug && len(origins) == 0 {
+		origins = append(origins, "*")
+	}
+	return origins
+}
```
yk-eukarya, line 146:
```suggestion
func allowedIntegrationOrigins(appCtx *ApplicationContext) []string {
	if appCtx == nil {
		return nil
	}
	origins := append([]string{}, appCtx.Config.IntegrationOrigins...)
	if appCtx.Debug {
		origins = append(origins, "http://localhost:3000", "http://127.0.0.1:3000", "http://localhost:8080")
	}
	return origins
}

func allowedPublicOrigins(appCtx *ApplicationContext) []string {
	if appCtx == nil {
		return nil
	}
	origins := append([]string{}, appCtx.Config.PublicOrigins...)
	if appCtx.Debug {
		origins = append(origins, "*")
	}
	return origins
}
```

yk-eukarya (approved):

PR #1539, opened 2025-06-30
fix(server): exported asset should include asset url

yk-eukarya:


mimoham24 (approved):
thank you, the refactoring looks very convenient

PR #1548, opened 2025-07-07
fix(server): Apply CORS after auth

jasonkarel:
# Overview
Apply CORS after auth to let all origin in integration api and public api
## How I tested
1. hit curl in browser with js script


yk-eukarya:
closed by #1858

PR #1563, opened 2025-07-24
feat(server): add visibility to create project params

nourbalaha:
# Overview
This PR adds visibility to create project params


yk-eukarya (approved):
LGTM

PR #1609, opened 2025-08-29
fix(server): asset upload check type

jasonkarel:
# Overview
implement data transfer check and asset upload check while creating asset


server/internal/usecase/interactor/asset.go
```diff
@@ -161,16 +161,38 @@ func (i *Asset) Create(ctx context.Context, inp interfaces.CreateAssetParam, op
 		var size int64
 		file = inp.File
 
+		visibility := project.VisibilityPrivate
```
yk-eukarya, line 164: By default, the project is public.

jasonkarel: ok

pyshx (approved):

PR #1643, opened 2025-09-29
fix(server): change bucket gcs redirect to endPoint with bucket and add debug logs

jasonkarel:
# Overview
GCS operations were failing with "InvalidBucketName" errors when using custom proxy endpoints. The bucket name was being corrupted from reearth-cms-dev-assets to just b, causing upload operations and asset decompression to fail.

Fixed custom endpoint format: Include bucket name in endpoint path (https://assets.cms.dev.reearth.io/reearth-cms-dev-assets)


server/internal/infrastructure/gcp/file.go
```diff
@@ -609,13 +615,47 @@ func getGCSObjectPathFolder(uuid string) string {
...
+
+	// For signed URLs, always use standard GCS client to ensure signatures work.
+	// with storage.googleapis.com, regardless of custom endpoint configuration
+	if forSigning || f.customEndpoint == "" {
+		log.Debugf("DEBUG: Using standard GCS client")
+		client, err = storage.NewClient(ctx)
+	} else {
+		// Include bucket name in the custom endpoint path to match expected format:
+		// https://assets.cms.dev.reearth.io/reearth-cms-dev-assets/...
+		endpointWithBucket := f.customEndpoint + "/" + f.bucketName
```
yk-eukarya, line 636: gcsproxy should append the bucket name if needed

server/internal/infrastructure/gcp/file.go
```diff
@@ -91,7 +95,8 @@ func (f *fileRepo) ReadAsset(ctx context.Context, u string, fn string, h map[str
 
 func (f *fileRepo) GetAssetFiles(ctx context.Context, u string) ([]gateway.FileEntry, error) {
 	p := getGCSObjectPath(u, "")
-	b, err := f.bucket(ctx)
+	// Use standard client for GetAssetFiles to avoid bucket name corruption
+	b, err := f.bucketWithOptions(ctx, true)
```
yk-eukarya, line 99: If we use the standard client here, the usage will not be tracked by the gcsproxy

PR #1645, opened 2025-09-30
feat(server): make database name configurable for migrations

nourbalaha:
# Overview
Makes the database name configurable for migrations via the `REEARTH_CMS_DB_CMS` environment variable, defaulting to `reearth_cms` if not set. This allows developers to use custom database names for local development without affecting the default behavior.

### DIFF
--- a/server/cmd/db-migrations/main.go
+++ b/server/cmd/db-migrations/main.go
@@ -49,7 +49,7 @@ func main() {
 	}
 
 	// get db name
-	dbName := os.Getenv("REEARTH_CMS_DB_NAME")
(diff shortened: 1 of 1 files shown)


server/cmd/db-migrations/main.go
```diff
@@ -48,10 +48,16 @@ func main() {
 		return
 	}
 
+	// get db name
+	dbName := os.Getenv("REEARTH_CMS_DB_NAME")
```
yk-eukarya, line 52: There is already an existing flag named `REEARTH_CMS_DB_CMS`, I think we can use it

PR #1688, opened 2025-10-27
feat(server): add topics field to project creation and updating logic [VIZ-2325]

wilfredmulenga:
-  Updates the Create and Update project APIs to handle topics, including validation and sentinel value handling for deletion.
-  In the gRPC request when updating a project, the `topics` field is an empty array even when its not explicitly passed as its not an optional field. So had to make a workaround to not update `topics` when it is an empty array but only when the array has an empty string. 

gRPC request to create a project with topics:
```
grpcurl -plaintext \
  -H "authorization: Bearer token" \
  -import-path server/schemas/internalapi/v1 \
  -proto schema.proto \
  -H "user-id: 01k6me4n1rnhsnvrrp1zweddz5" \
  -d '{"workspace_id": "01k6me4n1rnhsnvrrp22cv2xrw", "alias": "test10",  "name": "New Project", "topics":{ "values":["river"]}}' \
  localhost:50052 \
  reearth.cms.v1.ReEarthCMS/CreateProject
```

gRPC request to update a project with topics:
```
grpcurl -plaintext \
  -H "authorization: Bearer token" \
  -import-path server/schemas/internalapi/v1 \
  -proto schema.proto \
  -H "user-id: 01k6me4n1rnhsnvrrp1zweddz5" \
  -d '{ "project_id": "01k8mr1zfzg9abc2bm6ajhvrk4",  "topics":{"values":["history","gis","games"]}}' \
  localhost:50052 \
  reearth.cms.v1.ReEarthCMS/UpdateProject
```

### DIFF
--- a/server/internal/infrastructure/mongo/mongodoc/event_test.go
+++ b/server/internal/infrastructure/mongo/mongodoc/event_test.go
@@ -28,6 +28,10 @@ func TestEventDocument_Model(t *testing.T) {
 			Visibility: "public",
 		},
 	}
+
+	pDoc.StarredBy = []string{}
+	pDoc.Topics = []string{}
+
 	tests := []struct {
 		name    string
 		eDoc    EventDocument
@@ -83,6 +87,8 @@ func TestNewEvent(t *testing.T) {
 		Description: "xyz",
 		Alias:       "ppp123",
 		ImageURL:    "https://huho.com/xzy",
+		StarredBy:   []string{},
+		Topics:      []string{},
 		Workspace:   wId.String(),
 		Accessibility: &ProjectAccessibilityDocument{
 			Visibility: "public",


server/pkg/project/project.go
```diff
@@ -157,6 +162,14 @@ func (p *Project) Unstar(uId accountdomain.UserID) {
...
 }
 
+func (p *Project) SetTopics(topics []string) {
+	if p == nil {
+		return
+	}
+
+	p.topics = slices.Clone(topics)
+}
+
```
yk-eukarya: Topics should be validated, ie no duplication, no empty values ...

wilfredmulenga: Topics wont be case-sensitive, we had a discussion with Red about it and decided to not handle case-sensitivity. Thanks for considering it. 

I added e2e tests to handle duplicates and empty string values so it should cover the scenarios. Also reduces number of tests in the repo.

yk-eukarya (approved):

PR #1695, opened 2025-10-30
feat: [BREAKING CHANGE] update integration api endpoints to use workspaceIdOrAlias

yk-eukarya:
# Overview
Since the project alias will be unique under the workspace only (not on the entire db), the api should accept a workspace identifier with the project one.

## What I've done
I've refactored the integration api to maintain the following tree:
```
Workspaces
  Projects
    Models
      Items
        Comments
    Groups
    Assets
      Comments
```

### DIFF
--- a/server/internal/usecase/interactor/project_test.go
+++ b/server/internal/usecase/interactor/project_test.go
@@ -1291,6 +1291,7 @@ func TestProject_StarProject(t *testing.T) {
 			name:  "star project owned by another user",
 			seeds: project.List{p1.Clone()},
 			args: args{
+				wIdOrAlias: workspace.IDOrAlias(wid1.String()),
 				pIdOrAlias: project.IDOrAlias(p1.ID().String()),
 				operator:   otherUserOp,
 			},


web/src/components/organisms/Project/settings/hooks.ts
```diff
@@ -95,10 +95,11 @@ export default () => {
     async (alias: string) => {
       if (!alias) return false;
 
-      const response = await CheckProjectAlias({ variables: { alias } });
+      if (!workspaceId) throw new Error();
```
nourbalaha: Please add a descriptive error message for better debugging.

```suggestion
      if (!workspaceId) {
        throw new Error("Workspace ID is required to check project alias");
      }
```

web/src/components/organisms/Workspace/hooks.ts
```diff
@@ -169,10 +169,11 @@ export default () => {
     async (alias: string) => {
       if (!alias) return false;
 
-      const response = await CheckProjectAlias({ variables: { alias } });
+      if (!workspaceId) throw new Error();
```
nourbalaha: Please add a descriptive error message for better debugging.

```suggestion
      if (!workspaceId) {
        throw new Error("Workspace ID is required to check project alias");
      }
```

PR #1722, opened 2025-12-08
feat(server): run health check on server initialization

nourbalaha:
# Overview

Added automatic health check execution when the server starts, providing early detection of connectivity and permission issues before the server begins accepting requests.


yk-eukarya (approved):
LGTM

PR #1769, opened 2026-02-04
feat: add GraphQL SSE transport for subscriptions

yk-eukarya:
## Summary

Add Server-Sent Events (SSE) transport support for GraphQL subscriptions as an alternative to WebSocket connections.

### Changes

**Server (`server/internal/app/`)**
- Add SSE transport to the GraphQL handler with 10-second keep-alive ping interval
- Simplify route registration by using `api.Any` to handle all HTTP methods (POST for queries/mutations, GET for WebSocket/SSE)

**Web (`web/src/gql/`)**
- Add `graphql-sse` dependency for SSE client support
- Implement SSE link for Apollo Client with async authentication header support
- Refactor link splitting logic with named predicates for better readability
- Support fallback to WebSocket via `context.useWS` option for backward compatibility

### Motivation

SSE provides a simpler, HTTP-based alternative to WebSocket for GraphQL subscriptions that:
- **Works natively with HTTP/2** - Cloud Run uses HTTP/2 by default, and SSE leverages HTTP/2's multiplexing capabilities without requiring a protocol upgrade like WebSocket
- Works better with certain proxies and load balancers
- Has simpler connection semantics (no upgrade handshake)
- Maintains compatibility with existing WebSocket subscriptions via context flag

### Testing

- SSE subscriptions can be tested by calling a subscription without `useWS: true` in context
- WebSocket subscriptions still work when `{ context: { useWS: true } }` is passed

### DIFF
--- a/web/yarn.lock
+++ b/web/yarn.lock
@@ -13380,7 +13380,7 @@ graphql-language-service@5.5.0, graphql-language-service@^5.3.0:
     nullthrows "^1.0.0"
     vscode-languageserver-types "^3.17.1"
 
-graphql-sse@^2.6.0:
+graphql-sse@2.6.0:
   version "2.6.0"
   resolved "https://registry.yarnpkg.com/graphql-sse/-/graphql-sse-2.6.0.tgz#d3af7ac13ebb19fb7042808f535dc45c700bc64c"
   integrity sha512-BXT5Rjv9UFunjQsmN9WWEIq+TFNhgYibgwo1xkXLxzguQVyOd6paJ4v5DlL9K5QplS0w74bhF+aUiqaGXZBaug==


web/package.json
```diff
@@ -123,6 +123,7 @@
     "formik": "2.4.6",
     "graphiql": "3.7.1",
     "graphql": "16.12.0",
+    "graphql-sse": "^2.6.0",
```
kazettique: We prefer to use fixed version: `"2.6.0"`

web/src/gql/provider.tsx
```diff
@@ -92,20 +124,21 @@ const Provider: React.FC<Props> = ({ children }) => {
     },
   });
 
-  const httpLink = new HttpLink({ uri: endpoint, credentials: "include" });
+  const isSubscription = ({ query }: { query: unknown }): boolean => {
+    const definition = getMainDefinition(query as Parameters<typeof getMainDefinition>[0]);
```
kazettique: I updated type a bit.

```suggestion
  const isSubscription = (operation: ApolloLink.Operation): boolean => {
    const definition = getMainDefinition(operation.query);
```

web/src/gql/provider.tsx
kazettique: Add a function here to check http protocol in the browser, in order to decide when should we use websocket or server side event

```suggestion
function getHttpProtocol(): "HTTP/2" | "HTTP/3" | "HTTP/1.1" | "HTTP/1.0" | "Unknown" {
  const navEntry = performance.getEntriesByType("navigation")[0] as
    | PerformanceNavigationTiming
    | undefined;
  if (!navEntry) return "Unknown";

  switch (navEntry.nextHopProtocol) {
    case "h2":
      return "HTTP/2";
    case "h3":
      return "HTTP/3";
    case "http/1.1":
      return "HTTP/1.1";
    case "http/1.0":
      return "HTTP/1.0";
    default:
      return "Unknown";
  }
}
```

PR #1771, opened 2026-02-05
feat(web): update my integration code example

kazettique:
# Overview

- Update API endpoint in my integration code example due to backend API endpoint changed, please refer to #1695 
- Adjust style
  - Use `<pre />` to represent code content
  - Use `<Typography.Text code>` for highlight part, avoid using quotes (which is hard to read)
- Add copy button

## Memo

It's nice if we can get endpoint from backend (if possible)


nourbalaha (approved):
LGTM!

PR #1784, opened 2026-02-13
fix(server): Refactor Export Items

jasonkarel:
# Overview
Export Item Performance Optimization

  Problem
Export endpoints were experiencing severe performance degradation (15-25s for 1100+ items) due to N+1 query problems when loading referenced items. Each item with reference fields triggered individual database
  queries, resulting in thousands of queries per export.

  Cod changes: 

  Batch pre-loading of referenced items:
  - Collect all reference IDs from the current batch of items
  - Load all referenced items in a single query using FindByIDs
  - Store in memory map for O(1) lookup
  - ItemLoader uses pre-loaded cache instead of making database calls

  Privacy/security maintained:
  - IncloudRefModels filter applied at ItemLoader level
  - Only referenced items from allowed models are expanded
  - Items from filtered models show as ID strings (expected behavior)

  Exporter modification:
  - Changed JSONExporter to store request pointer instead of caching ItemLoader
  - Ensures per-batch ItemLoader updates are used correctly

  Files Changed

  - internal/usecase/interactor/item_export.go - Pre-load referenced items per batch
  - internal/adapter/publicapi/item.go - Batch load items, schemas, and assets
  - internal/adapter/integration/item.go - Batch load referenced items
  - pkg/exporters/json_exporter.go - Use request pointer for dynamic ItemLoader
  
## What I've done
1. code development
2. unit test verification
3. local testing
## How I tested
Tested public api and integration api and made sure there is no difference in responses

### DIFF
--- a/server/internal/usecase/interactor/item_export.go
+++ b/server/internal/usecase/interactor/item_export.go
@@ -99,7 +99,7 @@ func (i Item) Export(ctx context.Context, params interfaces.ExportItemParams, w 
 		} else {
 			refItemIDs = items.RefItemIDs(params.SchemaPackage)
 		}
-		var referencedItemsMap map[id.ItemID]*item.Item
+		var referencedItemsMap item.Map
 		if len(refItemIDs) > 0 {
 			versionedRefs, err := i.repos.Item.FindByIDs(ctx, refItemIDs, ver)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


server/internal/usecase/interactor/item_export.go
```diff
@@ -149,3 +206,26 @@ func (i Item) Export(ctx context.Context, params interfaces.ExportItemParams, w
...
+							seen[refID] = true
+						}
+					}
+				}
+			}
+		}
+	}
+
+	return refIDs
+}
```
yk-eukarya: Move this logic to the /pkg/item.list to be like:
```
func (l List) RefItemsIDs(sp schema.Package) ItemIDList {
	...
}
```

PR #1819, opened 2026-03-18
feat(server): integrate checkPermission cerbos in cms for model and project

jasonkarel:
# Overview

Add Cerbos-based permission checks (CheckPermission) to all public functions in model and project interactors, with unit tests covering allowed/denied/error cases. Fix data race in model tests by giving each parallel subtest its own model instances.


yk-eukarya (commented):
I think the implementation does not fully reflect the requirements. The user might have different roles on different workspaces, so the workspace ID should be passed to cerbos in order to do the check!

server/internal/usecase/interactor/project.go
```diff
@@ -433,6 +505,10 @@ func (i *Project) DeleteAPIKey(ctx context.Context, pId id.ProjectID, kId id.API
 	if err != nil {
 		return nil, err
 	}
+	pws3 := p.Workspace()
+	if err := i.checkPermission(ctx, op, &pws3, "project.DeleteAPIKey", rbac.ActionManageAPIKeys); err != nil {
+		return nil, err
+	}
```
yk-eukarya:
```suggestion
	if err := i.checkPermission(ctx, op, p.Workspace().Ref(), "project.DeleteAPIKey", rbac.ActionManageAPIKeys); err != nil {
		return nil, err
	}
```

yk-eukarya (commented):
How about the operator checks like `op.IsReadable...` ...etc, did you think what we should do about it?

server/internal/app/config.go
```diff
@@ -101,9 +101,10 @@ type HealthCheckConfig struct {
...
 
 type AccountAPIConfig struct {
-	Enabled bool   `pp:",omitempty"`
-	Host    string `pp:",omitempty"`
-	Timeout int    `pp:",omitempty"`
+	Enabled          bool   `pp:",omitempty"`
+	Host             string `pp:",omitempty"`
+	Timeout          int    `pp:",omitempty"`
+	Check_Permission bool   `pp:",omitempty"`
 }
```
yk-eukarya, line 108: update related .env.example

server/internal/usecase/interactor/project.go
```diff
@@ -56,29 +105,42 @@ func (i *Project) FindByWorkspaces(ctx context.Context, wIds accountdomain.Works
 	return i.repos.Project.Search(ctx, *f)
 }
 
-func (i *Project) Search(ctx context.Context, f interfaces.ProjectFilter, _ *usecase.Operator) (project.List, *usecasex.PageInfo, error) {
+func (i *Project) Search(ctx context.Context, f interfaces.ProjectFilter, operator *usecase.Operator) (project.List, *usecasex.PageInfo, error) {
+	var wid *workspace.ID
```
yk-eukarya: The user will be able to pass the workspaceIds as an empty parameter! But he will get results!
So we should check the permissions after the DB query 🤔

yk-eukarya: not resolved

server/internal/usecase/interactor/project.go
```diff
@@ -45,7 +88,13 @@ func (i *Project) FindByWorkspace(ctx context.Context, wid accountdomain.Workspa
...
 }
 
-func (i *Project) FindByWorkspaces(ctx context.Context, wIds accountdomain.WorkspaceIDList, f *interfaces.ProjectFilter, _ *usecase.Operator) (project.List, *usecasex.PageInfo, error) {
+func (i *Project) FindByWorkspaces(ctx context.Context, wIds accountdomain.WorkspaceIDList, f *interfaces.ProjectFilter, operator *usecase.Operator) (project.List, *usecasex.PageInfo, error) {
+	for _, ws := range wIds {
+		ws := ws
+		if err := i.checkPermission(ctx, operator, &ws, "project.FindByWorkspaces", rbac.ActionList); err != nil {
+			return nil, nil, err
+		}
+	}
```
yk-eukarya: unify the implementation, Project.Fetch(...) has same goal but defferent implementation!

server/internal/usecase/interactor/model.go
```diff
@@ -38,43 +40,146 @@ func NewModel(r *repo.Container, g *gateway.Container) interfaces.Model {
...
+	if len(models) > 0 {
+		ws, wsErr := i.workspaceIDForProject(ctx, models[0].Project())
+		if wsErr != nil {
+			return nil, wsErr
+		}
+		wid = ws.Ref()
+	}
+	if err := i.checkPermission(ctx, operator, wid, "model.FindByIDs", rbac.ActionList); err != nil {
+		return nil, err
+	}
```
yk-eukarya: Can we confirm that all models belong to the same project here? If not, we should check all of them!

server/internal/usecase/interactor/model.go
```diff
@@ -213,6 +329,17 @@ func (i Model) CheckKey(ctx context.Context, pId id.ProjectID, s string) (bool,
 }
 
 func (i Model) Delete(ctx context.Context, modelID id.ModelID, sp schema.Package, operator *usecase.Operator) error {
+	m, err := i.repos.Model.FindByID(ctx, modelID)
+	if err != nil {
+		return err
+	}
```
yk-eukarya, line 311: The model is fetched again inside the Run0(...), you can remove that call.

server/internal/usecase/interactor/model.go
```diff
@@ -387,6 +514,9 @@ func (i Model) deleteItemsByModel(ctx context.Context, prj *project.Project, m *
 }
 
 func (i Model) FindOrCreateSchema(ctx context.Context, param interfaces.FindOrCreateSchemaParam, operator *usecase.Operator) (*schema.Schema, error) {
+	if err := i.checkPermission(ctx, operator, nil, "model.FindOrCreateSchema", rbac.ActionRead); err != nil {
+		return nil, err
+	}
```
yk-eukarya: Don't forget to pass the workspace id

mimoham24 (approved):

PR #1852, opened 2026-04-09
feat(server): expose fields in access api: Created At, Created By, Updated At, UpdatedBy

jasonkarel:
# Overview
- Public API item and asset responses now include createdAt, updatedAt, createdBy, and updatedBy fields


server/pkg/exporters/map.go
```diff
@@ -170,5 +170,24 @@ func MapFromItem(itm *item.Item, sp *schema.Package, al AssetLoader, il ItemLoad
...
+
+	if itm.UpdatedByUser() != nil {
+		m["updatedBy"] = itm.UpdatedByUser().String()
+	} else if itm.UpdatedByIntegration() != nil {
+		m["updatedBy"] = itm.UpdatedByIntegration().String()
+	} else if itm.User() != nil {
+		m["updatedBy"] = itm.User().String()
+	} else if itm.Integration() != nil {
+		m["updatedBy"] = itm.Integration().String()
+	}
```
yk-eukarya: The user might have used the same keys! So agreed to use '$createdBy'. So each key has '$' as a prefix.

jasonkarel: i see, so later in long term solution, we should reserve them at schema level right ?

server/internal/adapter/publicapi/types.go
```diff
@@ -14,13 +15,25 @@ import (
...
 	m := i.Fields
 	m["id"] = i.ID
+	m["createdAt"] = i.CreatedAt
+	m["updatedAt"] = i.UpdatedAt
+	if i.CreatedBy != "" {
+		m["createdBy"] = i.CreatedBy
+	}
+	if i.UpdatedBy != "" {
+		m["updatedBy"] = i.UpdatedBy
+	}
```
yk-eukarya: You need to include the '$' here as well.

yk-eukarya (approved):

PR #1871, opened 2026-05-08
ci: grant explicit write permissions to labeler and assign-author jobs

yk-eukarya:
## Summary

- The org-wide default GitHub Actions token permissions were tightened to read-only, breaking the `labeler` and `assign-author` jobs on all PRs.
- Both jobs need write access to labels/assignees but were relying on the implicit default token permissions, which no longer include write access.
- Adds explicit `permissions` blocks to each job with the minimum required scopes (`pull-requests: write` for labeler, `issues: write` + `pull-requests: write` for assign-author).

## Test plan

- Open or update any PR and verify `labeler` and `assign-author` jobs pass.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -11,6 +11,9 @@ jobs:
   labeler:
     name: labeler
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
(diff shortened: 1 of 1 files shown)


yk-eukarya:
closed by #1868

PR #1884, opened 2026-05-26
feat(server): schema-based payload validation for POST endpoint

nourbalaha:
# Overview

Adds Schema.ValidatePayload() in pkg/schema to validate raw key→value payloads against the model schema. Wires it into the PostItem handler, returning field-level errors with machine-readable codes. Asset, reference, tag, and group fields are out of scope for phase 1.


server/pkg/schema/fields_validation_test.go
yk-eukarya, line 1: It's more readable if the test cases are written in table-driven tests

yk-eukarya (approved):
LGTM

PR #1894, opened 2026-06-02
feat(server): Model-level posting enabled flag (postingEnabled)

jasonkarel:
# Overview
Model-level posting enabled flag (postingEnabled)

Adds an independent postingEnabled flag to the model domain that gates the public API POST endpoint at the model level, separate from the existing project-level posting config.

What changed
- Domain (pkg/model) — added postingEnabled bool field with getter, setter, and builder method. Defaults to false.
- GQL schema — new ModelPostingSettings type and UpdateModelPostingSettingsInput input, both independent from the project-level PostingSettings. Exposed as postingSettings on the Model type and UpdateModelInput.
- MongoDB — PostingEnabled bool persisted on the model document. Missing field in existing documents reads back as false (safe default).
- Public API (ValidatePostingAccess) — the POST endpoint now checks model posting enabled in addition to project posting enabled. Both must be true for a request to proceed.

GraphQL
Read current state

```
query {
  node(id: "<modelId>", type: Model) {
    id
    ... on Model {
      postingSettings {
        enabled
      }
    }
  }
}
```

Enable posting on a model

```
mutation {
  updateModel(input: {
    modelId: "<modelId>"
    postingSettings: { enabled: true }
  }) {
    model {
      id
      postingSettings {
        enabled
      }
    }
  }
}
```

Disable posting on a model
```
mutation {
  updateModel(input: {
    modelId: "<modelId>"
    postingSettings: { enabled: false }
  }) {
    model {
      id
      postingSettings {
        enabled
      }
    }
  }
}
```


server/internal/adapter/publicapi/item_test.go
yk-eukarya, line 1: `modelPostingEnabled == false` is not covered.

server/internal/adapter/publicapi/item_test.go
```diff
@@ -97,47 +104,43 @@ func TestController_PostItem(t *testing.T) {
...
+		wantFieldErrors bool
 	}{
 		{
-			name:           "posting disabled returns ErrProjectPostingDisabled",
-			postingEnabled: false,
-			allowedOrigins: []string{allowedOrigin},
-			wantErr:        ErrProjectPostingDisabled,
+			name:    "valid body returns no error",
+			wantErr: nil,
 		},
```
yk-eukarya, line 123: empty body should be an invalid case!

server/internal/adapter/publicapi/item_test.go
```diff
@@ -149,41 +152,68 @@ func TestController_PostItem(t *testing.T) {
...
+			result := ctrl.PostItem(ctx, wAlias, pAlias, mKey, body)
 
 			if tt.wantErr != nil {
 				assert.ErrorIs(t, result.Err, tt.wantErr)
 			} else {
 				assert.NoError(t, result.Err)
 			}
+			if tt.wantFieldErrors {
+				assert.NotEmpty(t, result.FieldErrors)
+			}
```
yk-eukarya: assert.NotEmpty() is not enough, plz use assert.ErrorIs()

yk-eukarya (approved):

server/internal/adapter/publicapi/item_test.go
jasonkarel: i changed it to struct and used equal for more precise check.

PR #1905, opened 2026-06-11
feat(server): draft item creation via Access API

nourbalaha:
# Overview

Wires the validated Access API payload into the existing item creation usecase, enforcing Draft status on all submissions.

- Introduces a `Posting` operator type (no user/integration identity) scoped to the target project, allowing the existing `item.Create` usecase to be reused without a parallel path
- Bypasses the workspace write check for posting operators; replaces it with a project-scoped check
- Maps validated body fields to `CreateItemParam` via schema key lookup
- Returns `{ id, $createdAt, fields }` on success (202)
- Unit tests: field mapping logic, Draft status verified against the persisted item, handler response shape


server/pkg/item/builder.go
```diff
@@ -114,6 +114,11 @@ func (b *Builder) IsMetadata(im bool) *Builder {
 	return b
 }
 
+func (b *Builder) Anonymous(v bool) *Builder {
+	b.i.isAnonymous = v
```
yk-eukarya, line 125: we can validate the item in the builder.Build that has a user, intigration or its an anonymous item.

server/internal/app/public_api.go
```diff
@@ -0,0 +1,43 @@
...
+					strings.HasSuffix(c.Request().URL.Path, "/items")
+			},
+		}))
+
+		// register dummy OPTIONS route so CORS middleware works fine!
+		publicAPIGroup.OPTIONS("/*", func(ctx *echo.Context) error { return nil })
+	}
+
+	publicAPIGroup.Use(publicAPIAuthMiddleware(appCtx), usecaseMiddleware)
+	publicapi.Echo(publicAPIGroup, PublicAPIPostingMiddleware())
```
yk-eukarya: I think we can apply the anonymous operator for all public api 🤔

server/internal/adapter/publicapi/api.go
```diff
@@ -257,6 +258,24 @@ type postItemRequest struct {
 	Fields map[string]any `json:"fields"`
 }
 
+// postingAccessError writes the appropriate HTTP error response for known posting-access errors
+// and returns true if the error was handled. Returns false for unknown errors.
+func postingAccessError(c *echo.Context, err error) (bool, error) {
```
yk-eukarya: Errors are already refactored in pr #1908

server/internal/adapter/publicapi/item.go
```diff
@@ -165,31 +167,87 @@ func getReferencedItems(ctx context.Context, i *item.Item, sp *schema.Package, p
 	return vi
 }
 
+type PostItemResponse struct {
```
yk-eukarya, line 169: the response type should be defiend in the public api oapi schema as well

server/internal/adapter/publicapi/item.go
```diff
@@ -165,31 +167,87 @@ func getReferencedItems(ctx context.Context, i *item.Item, sp *schema.Package, p
...
+		v, ok := body[key.String()]
+		if !ok {
+			continue
+		}
+		// itemFieldsFromParams expects []any for multiple fields; normalize scalars here.
+		if f.Multiple() {
+			if _, isSlice := v.([]any); !isSlice {
+				v = []any{v}
+			}
+		}
```
yk-eukarya: shouldn't we return a validation error?

server/e2e/publicapi_test.go
```diff
@@ -1255,6 +1256,24 @@ func TestPublicAPI_PostItem(t *testing.T) {
 		})
 	})
 
+	t.Run("response fields reflect submitted values", func(t *testing.T) {
```
yk-eukarya, line 1259: add another test that assert the created item is not published

server/internal/app/public_api.go
```diff
@@ -0,0 +1,61 @@
...
+
+		// register dummy OPTIONS route so CORS middleware works fine!
+		publicAPIGroup.OPTIONS("/*", func(ctx *echo.Context) error { return nil })
+	}
+
+	publicAPIGroup.Use(publicAPIAuthMiddleware(appCtx), usecaseMiddleware, PublicAPIPostingMiddleware())
+	publicapi.Echo(publicAPIGroup)
+}
+
+func PublicAPIPostingMiddleware() echo.MiddlewareFunc {
```
yk-eukarya:
```suggestion
func AnonymousOperatorMiddleware() echo.MiddlewareFunc {
```

server/internal/adapter/publicapi/item_test.go
```diff
@@ -495,3 +495,74 @@ func TestHandler_PreflightItem(t *testing.T) {
...
+		Project(id.NewProjectID()).
+		Fields([]*schema.Field{textField, numberField}).
+		MustBuild()
+
+	tests := []struct {
+		name       string
+		body       map[string]any
+		wantKeys   []string
+		wantAbsent []string
+	}{
```
yk-eukarya, line 523:
```suggestion
	tests := []struct {
		name       string
		input      map[string]any
		want       []interfaces.ItemFieldParam
	}{
```

server/internal/adapter/publicapi/openapi.go
```diff
@@ -227,6 +205,53 @@ func (c *Controller) GetOpenAPISchema(ctx context.Context, wsAlias, pAlias strin
 				},
 			},
 		}
+		if postingEnabled && m.PostingEnabled() {
+			itemsPath["post"] = map[string]interface{}{
+				"summary":     fmt.Sprintf("Create a new %s item", m.Name()),
+				"description": fmt.Sprintf("Submit a new item to the %s model", m.Name()),
+				"requestBody": map[string]interface{}{
```
yk-eukarya, line 212: it more clear to mention that the item is unpublished

server/internal/adapter/publicapi/item.go
nourbalaha: I think no, `fieldsFromBody` is the conversion step that runs after `ValidateFields` has already passed. By the time we get here, we know the value is valid. The `ValidateFields` function already catches the reverse case (passing an array to a single field):

```
if _, isSlice := raw.([]any); isSlice && !f.Multiple() {
    return &FieldValidationError{Field: key, Code: FieldValidationCodeTypeMismatch}
}
```

The normalization in `fieldsFromBody` handles a convenience case: if someone sends "tags": "foo" instead of "tags": ["foo"] for a multiple field, we wrap it silently. `ValidateFields` doesn't reject this, it only rejects arrays on single fields, not scalars on multiple fields. So I think wrapping it here is the right behavior, not an error.

yk-eukarya (approved):

PR #1908, opened 2026-06-12
feat(server): access api cors enforcement

jasonkarel:
# Overview

Removes the user-facing enabled flag from project posting settings and enforces per-project CORS on the public posting endpoint (POST /api/p/:workspace/:project/:model/items).

Changes
Posting enabled flag

- Removed from the GraphQL schema — posting is now always on at the project level and can't be toggled via the API.
- Kept internally as a Mongo-only kill switch: defaults to true (missing field = enabled), can be disabled by setting accessibility.posting.enabled: false directly in the DB. No migration; model-level flag unaffected.

CORS enforcement (TI-2)

- Browser requests (Origin present) are validated against the project's allowedOrigins: empty list or mismatch → 403 origin_not_allowed; match → exact origin echoed in Access-Control-Allow-Origin (never *).
- Preflight: 204 with Allow-Methods: POST, Allow-Headers: Content-Type (no Max-Age); rejected → 403.
- Non-browser clients (no Origin) skip CORS entirely.
- Runs before posting logic; existing public API endpoints unaffected.

Misc

- Error codes/messages and the shared error→response mapper moved to a new publicapi/error.go; added isBrowserRequest helper.
- Unit + e2e tests cover all CORS paths and the kill-switch default; full test suite passes.
- Testing details: https://app.notion.com/p/eukarya/C-A1-001-WP4-T1-36616e0fb16580a5bd18ff32ffbb8b59?source=copy_link


server/pkg/project/posting_settings_test.go
yk-eukarya, line 137: keep the test for both cases

server/internal/infrastructure/mongo/mongodoc/project_test.go
```diff
@@ -247,3 +247,51 @@ func TestProjectPublicationDocument_Model(t *testing.T) {
...
+			require.NoError(t, err)
+			if tt.wantNil {
+				assert.Nil(t, got)
+				return
+			}
+			require.NotNil(t, got)
+			assert.Equal(t, tt.wantEnabled, got.Enabled())
+			assert.Equal(t, tt.wantOrigins, got.AllowedOrigins())
+		})
+	}
```
yk-eukarya, line 299:
```suggestion
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			got, err := tt.doc.Model()
			if tt.wantErr {
				assert.ErrorIs(t, err, tt.wantErr)
				assert.Nil(t, got)
				return
			}
			assert.NoError(t, err)
			assert.Equal(t, tt.want, got)
		})
	}
```

server/internal/adapter/publicapi/error.go
```diff
@@ -0,0 +1,66 @@
...
+			Message: msgOriginNotAllowed,
+		})
+	case errors.Is(err, rerror.ErrNotFound):
+		return c.JSON(http.StatusNotFound, apiErrorResponse{
+			Error: "not found",
+		})
+	default:
+		return err
+	}
+}
```
yk-eukarya, line 79: The errors aren’t consistent; some have messages while others don’t.  Furthermore, I believe the error code should be defined in the open API schema as well.

server/internal/adapter/publicapi/item.go
```diff
@@ -201,5 +204,8 @@ func (c *Controller) ValidatePostingAccess(ctx context.Context, wsAlias, pAlias,
 	if !wpm.Model.PostingEnabled() {
 		return ErrModelPostingDisabled
 	}
+	if !isBrowserRequest(origin) {
```
yk-eukarya, line 207: shouldn't we block the request in this case?

jasonkarel: the reason it passes through is because the requirement says "Non-browser clients are unaffected."

server/internal/adapter/publicapi/openapi.go
```diff
@@ -92,6 +137,49 @@ func (c *Controller) GetOpenAPISchema(ctx context.Context, wsAlias, pAlias strin
...
+						},
+						"404": map[string]interface{}{
+							"description": "Workspace, project, or model not found (not_found)",
+							"content":     errorContent,
+						},
+					},
+				},
+			}
+		}
+
```
yk-eukarya, line 182: This endpoint will draft an item, or it will create an item in draft state.

yk-eukarya (approved):

PR #1914, opened 2026-06-22
fix(server): resolve anonymous items in GQL and cover origin marker with tests

nourbalaha:
# Overview

Anonymous items created via the Access API (no user, no integration) failed to
load in the CMS. The `itemResolver.CreatedBy` GQL resolver dereferenced a nil
`IntegrationID` for these items, panicking the resolver and breaking the item
query.

- Guard `CreatedBy` to return `nil` when an item has neither user nor
  integration, mirroring the existing `UpdatedBy` behavior. `createdBy` is
  nullable in the schema, so no frontend change is needed.
- Add unit tests for the `isAnonymous` origin marker: domain getter/`Clone`/
  default-false (`pkg/item`) and the mongodoc persistence round-trip both
  directions, closing the WP3-T4 testing gap.


yk-eukarya (approved):
LGTM

PR #1917, opened 2026-06-24
feat(server): rate limit the public posting endpoint

nourbalaha:
## Overview

Adds per-IP rate limiting to the public posting endpoint only.


server/internal/app/config.go
```diff
@@ -85,6 +86,18 @@ type Config struct {
 	Policy_Checker PolicyCheckerConfig `pp:",omitempty"`
 }
 
+const (
+	defaultPublicRateLimitPerMinute = 100
+	defaultPublicRateLimitBurst     = 100
+	defaultPublicRateLimitExpires   = 1 * time.Minute
+)
+
+type PublicRateLimitConfig struct {
```
yk-eukarya:
```suggestion
type RateLimitConfig struct {
```

nourbalaha: Fixed!

yk-eukarya (approved):

PR #799, opened 2023-08-23
feat(server): publish multiple items at once

nourbalaha:
# Overview
- Publish multiple items instead of one item
- Rename itemId to itemIds in `UnpublishItemInput`

PR #1921, opened 2026-06-30
fix(server): reduce per-item DB round-trips in publish/unpublish

jasonkarel:
# Overview

I chose this problem from AI scanning:
🟠 [SCA-02] [HIGH] PublishItem / UnpublishItem: per-item DB writes + webhook dispatch inside one transaction
Location: [server/internal/usecase/interactor/item.go:548-581](https://github.com/reearth/reearth-cms/blob/b36634dcb00adc57daf90cca4c0c9e2b8b752c5a/server/internal/usecase/interactor/item.go#L548-L581)
Introduced: [b21588d](https://github.com/reearth/reearth-cms/commit/b21588d9a9046684a40971aceb7ffd016ac35f72) · Nour Balaha · 2023-08-23 · "feat(server): publish multiple items at once (#799)"
Problem: The GraphQL publishItem/unpublishItem mutations accept a user-supplied ItemIds list with no cap and loop over it calling Item.UpdateRef per item, then loop again calling getReferencedItems (a FindByIDs) plus i.event per item — and createEvent itself runs Workspace.FindByID + Integration.FindByIDs + TaskRunner.Run per item, all wrapped in a single Run1(...Transaction...). (Unpublish: same pattern at item.go:481-508.)
Risk: ~5 DB/external calls × N items in one long transaction → holds DB locks and floods the task queue; a single request publishing thousands of items serializes thousands of round-trips and degrades throughput under concurrent mutations.

Changes

- Batched ref update: replaced the per-item UpdateRef loop with a single BulkUpdateRef write (added to the item repository + Mongo/memory implementations).
- Batched referenced-items fetch: the per-item FindByIDs calls are now resolved in one query for the whole batch, then mapped back per item so each webhook payload keeps only its own referenced items.
- Deferred event/webhook dispatch: events are collected during the transaction and dispatched after it commits, so TaskRunner.Run, Workspace.FindByID, and Integration.FindByIDs no longer run inside the transaction. The batched dispatch also does a single shared workspace/integration lookup instead of one per item.


server/internal/usecase/interactor/item.go
```diff
@@ -819,6 +845,11 @@ func (i Item) events(ctx context.Context, e []Event) error {
 }
 
 func (i Item) getReferencedItems(ctx context.Context, fields []*item.Field) ([]item.Versioned, error) {
+	return i.repos.Item.FindByIDs(ctx, referencedItemIDs(fields), nil)
+}
+
+// referencedItemIDs collects the unique referenced item ids from the given fields.
+func referencedItemIDs(fields []*item.Field) id.ItemIDList {
```
yk-eukarya: plz use the `RefItemIds` in `Item` and `ItemList`

server/internal/usecase/interactor/item.go
```diff
@@ -829,10 +860,42 @@ func (i Item) getReferencedItems(ctx context.Context, fields []*item.Field) ([]i
...
+	refs, err := i.repos.Item.FindByIDs(ctx, allIDs, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	byID := make(map[id.ItemID]item.Versioned, len(refs))
+	for _, r := range refs {
+		byID[r.Value().ID()] = r
+	}
+
```
yk-eukarya: use `ToMap` from `ItemList`

server/internal/usecase/interactor/item.go
```diff
@@ -818,21 +862,66 @@ func (i Item) events(ctx context.Context, e []Event) error {
...
+	gsm := lo.SliceToMap(groups, func(g *group.Group) (id.GroupID, *schema.Schema) {
+		return g.ID(), sl.Schema(lo.ToPtr(g.Schema()))
+	})
+	rs := lo.Map(s.ReferencedSchemas(), func(sid schema.ID, _ int) *schema.Schema {
+		return sl.Schema(&sid)
+	})
+
+	return schema.NewPackage(s, meta, gsm, rs), nil
+}
```
yk-eukarya: This code is duplicated. Check: https://github.com/reearth/reearth-cms/blob/5878e7d6d3de6e721d1b7acd53556c1aaca93942/server/internal/usecase/interactor/schema.go#L40

yk-eukarya (approved):

yk-eukarya (approved):

PR #1929, opened 2026-07-02
fix(server): harden public posting API

nourbalaha:
# Overview

Fixes correctness and error-handling issues found while reviewing the C-A1-001 Access API epic:

- Retry-After accuracy: ceil(1/rate) → ceil(burst/rate), floored at 1, so the header reflects actual burst-drain time instead of a misleadingly short retry hint
- Eliminates the redundant WPM load on every POST: ValidatePostingAccess now returns the loaded context so PostItem reuses it instead of re-fetching
- Maps ErrOperationDenied/ErrInvalidOperator to a clean 403 access_denied instead of falling through to a raw 500
- ErrNoOriginsConfigured now maps explicitly to 403 posting_disabled instead of a 500
- Cross-project posting now 404s instead of silently succeeding (model must belong to the project in the URL)
- SaveDraft enforces the same CanWrite check as Save in both memory and mongo repos, with a note on where project-scoping actually happens for the anonymous path
- Various e2e coverage additions (bad-origin scheme mismatch, malformed JSON against rate limit, unsupported field types, empty required fields)


server/internal/infrastructure/mongo/item.go
```diff
@@ -213,6 +213,9 @@ func (r *Item) Save(ctx context.Context, item *item.Item) error {
 }
 
 func (r *Item) SaveDraft(ctx context.Context, item *item.Item) error {
+	if !r.f.CanWrite(item.Project()) {
```
yk-eukarya: SaveDraft should not have a permission check

server/internal/infrastructure/memory/item.go
```diff
@@ -181,6 +181,9 @@ func (r *Item) SaveDraft(_ context.Context, t *item.Item) error {
 	if r.err != nil {
 		return r.err
 	}
+	if !r.f.CanWrite(t.Project()) {
```
yk-eukarya: SaveDraft should not have a permission check

server/internal/usecase/interactor/item.go
```diff
@@ -193,7 +196,11 @@ func (i Item) Create(ctx context.Context, param interfaces.CreateItemParam, oper
 			return nil, err
 		}
 
-		if !operator.Anonymous && !operator.IsWritableWorkspace(s.Workspace()) {
+		if operator.Anonymous {
+			if m.Project() != param.ProjectID {
```
yk-eukarya: this is already checked in the adapter layer

server/internal/infrastructure/mongo/item.go
nourbalaha: Fixed!

server/internal/infrastructure/memory/item.go
nourbalaha: Fixed!

server/internal/usecase/interactor/item.go
nourbalaha: Fixed!

yk-eukarya (approved):

PR #1931, opened 2026-07-07
test(server): add shared repo test suite and unify item repo tests

jasonkarel:
# Overview
Scope: the full test-case inventory, the new internal/usecase/repo/test package, the shared item suite, entry points in both backends, and removal of the old item tests.

- add internal/usecase/repo/test with an interface-level suite for repo.Item that both the mongo and memory backends run via thin entry-point tests
- fix memory Item.Archive to check write permission against the passed project ID, matching the mongo implementation and interface contract
- fix memory Item.Search to match all items when the keyword is empty
- keep genuinely backend-specific tests, clearly named: Copy filter format and archive BSON document (mongo), SetItemError injection and remove error semantics (memory)


server/internal/usecase/repo/test/item_test_suite.go
```diff
@@ -0,0 +1,733 @@
...
+		},
+	}
+
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			r := newRepo(t)
+			assert.NoError(t, r.Save(ctx, i1))
```
yk-eukarya:
```suggestion
			require.NoError(t, r.Save(ctx, i1))
```

server/internal/usecase/repo/test/item_test_suite.go
```diff
@@ -0,0 +1,733 @@
...
+
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			r := newRepo(t)
+			for _, i := range []*item.Item{i1, i2} {
+				assert.NoError(t, r.Save(ctx, i))
+			}
```
yk-eukarya:
```suggestion
			require.NoError(t, r.SaveAll(ctx, item.List{i1, i2}))
```

yk-eukarya (approved):

mimoham24 (approved):

PR #1941, opened 2026-07-15
fix(cross-tenant asset access): verify asset belongs to requested project in public API

franzramadhan:
## Problem

`GetAsset` in the public API validated that the URL project has public assets enabled, but never checked that the requested asset ID actually belongs to that project. `FindByID` runs with a `nil` operator (unscoped), so any asset in the system was reachable via a URL pointing at any public-assets-enabled project.

An attacker could point the URL at any project with public assets enabled and supply an asset ID belonging to a private project in a different workspace/tenant to read its filename, size, and access URL.

The internal API and m2m handler already perform the equivalent `asset.Project() == wpm.Project.ID()` check; this path omitted it.

## Fix

Add a project ownership assertion immediately after the `FindByID` call. Requests for an asset that does not belong to the URL project return 404.

## Risk category

A01/Broken Access Control (HIGH)

### DIFF
--- a/server/internal/usecase/interactor/item.go
+++ b/server/internal/usecase/interactor/item.go
@@ -258,7 +258,13 @@ func (i Item) Create(ctx context.Context, param interfaces.CreateItemParam, oper
 		return nil, err
 	}
 
-	return Run1(ctx, operator, i.repos,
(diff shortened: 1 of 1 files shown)


franzramadhan:

PR #1942, opened 2026-07-15
fix(SSRF asset from URL): block private/reserved IPs and non-http(s) schemes

franzramadhan:
## Problem

`file.FromURL` fetched any attacker-controlled URL server-side with no host filtering. An authenticated user could submit `url=http://169.254.169.254/latest/meta-data/iam/security-credentials/…` (AWS IMDSv1) or any internal service endpoint; the fetched response body would be stored as an asset in a project the user controls, exfiltrating internal/cloud-metadata responses.

Reachable from the GraphQL `createAsset` mutation and the integration `AssetCreate` endpoint. The only recent hardening added a request timeout for DoS, not SSRF filtering.

## Fix

- Replace the default HTTP transport with a custom `DialContext` (`ssrfSafeDialContext`) that resolves each hostname to IPs and rejects connections to private/link-local/reserved ranges (RFC 1918, loopback, AWS/GCP/Azure IMDS link-local, shared address space) on every dial attempt — including after redirects.
- Reject non-`http`/`https` schemes (`gopher://`, `file://`, etc.) before the request is built.

## Risk category

A10/Insecure Design — SSRF (HIGH)

### DIFF
--- a/server/internal/usecase/interactor/item.go
+++ b/server/internal/usecase/interactor/item.go
@@ -258,7 +258,13 @@ func (i Item) Create(ctx context.Context, param interfaces.CreateItemParam, oper
 		return nil, err
 	}
 
-	return Run1(ctx, operator, i.repos,
+	// ev is populated inside the transaction and published after it commits,
+	// mirroring the pattern already used by Asset.Create (asset.go:521).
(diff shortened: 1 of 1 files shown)


franzramadhan:

PR #1943, opened 2026-07-15
fix(S3 streaming upload): replace io.ReadAll+PutObject with chunked multipart upload

franzramadhan:
## Problem

`fileRepo.Upload` called `io.ReadAll` to materialize the entire file body into a `[]byte` before `PutObject`. On a 512 MB Cloud Run container a single upload of a few hundred MB exhausts available RAM; under default concurrency (80) concurrent uploads multiply the footprint. The GCS backend already streams via `io.Copy`; the S3 backend lacked the equivalent.

## Fix

Replace `io.ReadAll + PutObject` with the S3 multipart upload API (`CreateMultipartUpload / UploadPart / CompleteMultipartUpload`). The body is read in 8 MiB chunks so peak memory per upload is O(8 MiB) regardless of object size. Zero-byte objects fall back to a plain `PutObject` (S3 requires at least one part ≥ 5 MiB in multipart mode). On any error the multipart upload is aborted to avoid orphaned parts accruing storage charges.

No new dependencies required — uses the existing `github.com/aws/aws-sdk-go-v2/service/s3` APIs already imported.

## Risk category

SCA-01/Critical — OOM on large asset uploads

### DIFF
--- a/server/internal/usecase/interactor/item.go
+++ b/server/internal/usecase/interactor/item.go
@@ -258,7 +258,13 @@ func (i Item) Create(ctx context.Context, param interfaces.CreateItemParam, oper
 		return nil, err
 	}
 
-	return Run1(ctx, operator, i.repos,
+	// ev is populated inside the transaction and published after it commits,
+	// mirroring the pattern already used by Asset.Create (asset.go:521).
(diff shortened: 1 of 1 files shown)


franzramadhan:

PR #1944, opened 2026-07-15
fix(model delete event accumulation): flush threads and events per page

franzramadhan:
## Problem

`deleteItemsByModel` paginated through all items but accumulated every item's `Event` struct (embedding the full item, model, and schema) and every thread ID into two slices that grew for the duration of the entire model. For a model with many items this built an unbounded in-memory collection before any flushing occurred, potentially OOM-ing the container.

## Fix

Restructure to flush each page's thread IDs (`Thread.RemoveByIDs`) and events (`createEvents`) before advancing to the next page. Peak memory is now O(pageSize=100) regardless of total item count. The bulk `RemoveByModel` call for item deletion is unchanged — it remains a single DB operation at the end.

## Risk category

SCA-05/High — OOM during model deletion with large item counts

### DIFF
--- a/server/internal/usecase/interactor/item.go
+++ b/server/internal/usecase/interactor/item.go
@@ -258,7 +258,13 @@ func (i Item) Create(ctx context.Context, param interfaces.CreateItemParam, oper
 		return nil, err
 	}
 
-	return Run1(ctx, operator, i.repos,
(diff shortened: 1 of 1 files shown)


franzramadhan:

PR #1966, opened 2026-07-31
feat(web): posting api UI improvement

kazettique:
# Overview

Improve the Public API settings UI and add a per-model **Copy cURL** for the Posting API.

## What I've done

- Add `GET` / `POST` method tags to endpoints in the Reading & Posting tables
- Fix the Posting endpoint to append `/items` (matches `POST /:workspace/:project/:model/items`)
- Add a **Copy cURL** button per model — copies a multi-line `curl` with a request body built from the model's schema
  - Placeholders show each field's type + constraints (required, options, min / max, maxLength)
  - Only server-supported field types are included (Asset / Reference / Tag / Group / Geometry excluded)
- Mark the Posting tab as experimental (icon)
- Extend the `GetModels` query to fetch schema field detail (`key`, `required`, `multiple`, `typeProperty`)

## How I tested

- Manually verified the copied `curl` in the Posting tab

- `GetModels` selection is widened (shared by 5 consumers); the change is additive only


nourbalaha (approved):

mimoham24:
@kazettique 
Some optional follow-up suggestions for future improvement (non-blocking for this PR):

1. **Models with required unsupported fields:** The cURL generator omits unsupported field types, but if one of those fields is required, the server rejects posting for the model through `hasUnsupportedRequiredTypes`. As a future improvement, we could disable “Copy cURL” for such models or show a clear warning that Public API posting is not available for them. A test covering a required unsupported field would also be useful.

2. **Preserve JSON value types in examples:** Integer, Number, and Bool placeholders are currently emitted as quoted strings, although the API expects actual numbers and booleans. Later, we could generate valid sample values such as `1` and `false`, and display constraints separately from the value. Tests for Bool/Integer/Number would help here.

These are suggestions only and are not required for the current merge.

PR #1967, opened 2026-08-03
fix(server): stream and batch large asset file lists on decompress finalize

nourbalaha:
# Overview

Extracting archives with hundreds of thousands of files (e.g. 577k) could leave the archive extraction status stuck at "in progress" — the decompressor finished, but the server's finalize step (list all extracted files from storage, save to DB) ran unbounded and took long enough to exceed the request/ack timeout before the status could be updated to "Done".

- GetAssetFiles now streams entries via a callback instead of returning one full in-memory slice, avoiding double-buffering hundreds of thousands of FileEntry/asset.File objects at once.
- SaveFlat now sends its Mongo BulkWrite in batches instead of one unbounded call for all files.
- Added logging at each step of UpdateFiles and around the listing/bulk-write loops, so a slow or stuck finalize is diagnosable from logs (which step it reached, how many files/batches processed).


server/internal/usecase/interactor/asset_test.go
yk-eukarya, line 1: It's better to use the mock imp or memory imp.

mimoham24 (approved):

PR #2002, opened 2026-09-15
test(web): integrate Allure reporting for e2e tests

waqaszulfiqar:
# Overview

## What I've done
- Add allure-playwright reporter, allurerc.mjs, and report scripts
- Upload/publish Allure results in e2e web CI workflows
- Move DATA_TEST_ID and Test into src/test/data.ts so e2e no longer pulls @testing-library/react into Playwright's Node loader (crashed collection)


.github/workflows/e2e_web.yml
```diff
@@ -61,9 +64,14 @@ jobs:
...
     name: playwright
     runs-on: ubuntu-24.04-arm
+    concurrency:
+      group: e2e-web-allure-reports
+      cancel-in-progress: false
     permissions:
       contents: read
       id-token: write
+    env:
+      BUCKET: reearth-oss-cms-allure-reports
```
nourbalaha: The bucket name is hardcoded here. If we ever need a different bucket (like for staging), we'd have to edit this file directly instead of just changing a variable. Since this file already uses vars.* for other settings, it'd be more consistent to do the same here, e.g. ${{ vars.ALLURE_BUCKET || 'reearth-oss-cms-allure-reports' }}.

nourbalaha (approved):

kazettique (approved):

mimoham24 (approved):

pyshx (approved):

PR #2006, opened 2026-09-17
feat(server): add Japanese translations of OpenAPI specs

nourbalaha:
# Overview

Adds `publicapi.ja.yaml` and `integration.ja.yml`, Japanese versions of the Public API and Integration API OpenAPI specs, for use by the docs.reearth.io documentation site. Only human-readable text is translated.


asahina820 (approved):
I checked Japanese translation side, LGTM!

